PR #3, opened 2016-02-05
Add colDeSac fixture

ingalls:
screenshot 2016-02-05 05 08 51

No match is generated for this road segment and corresponding points


ingalls:
The address points in this instance didn't have a `number` field.

PR #41, opened 2016-09-17
Self Check

ingalls:
- Use raw addresses from address input mbtiles to test how percentage of addresses that are encoded into the output itp layer. `test` mode.

- Also enforces numeric (Integer) addresses. **Backwards Incompatible**


ingalls:
Ref: https://github.com/mapbox/carmen/issues/496

PR #49, opened 2016-10-19
Turtles all the way down

ingalls:
## Current Problem

Although the tile reduce library is crazy fast and awesome it has one glaring problem that has caused large improvements to the core algorithm difficult. This is that streets and their associated addresses can become orphaned from each other when separated by a tile boundary. 

Example:

```
A B
```

Tiles `A` and `B` where `A` contains the road geometry and `B` contains the address points. Since no data is shared across tiles, they can never be matched to each other.
## New Approach

heh, we'll figure that as we go along. Current plan is first grouping streets/addresses by name into files on disk and then parsing each cluster by geographic proximity. Then a second pass would do the same for unnamed streets.

Will update this section as I work out the details.

cc/ @yhahn @sbma44


sbma44:
makes sense, though I should ask: have you looked at generating tiles with overlap? that's how I'm handling this in post.dedupe: https://github.com/mapbox/mapbox-places/blob/wikidata-landmark-dedupe/post.dedupe/script.sh#L34

PR #78, opened 2017-05-11
Freq Match => Lev. Distance

ingalls:
- :bug: Allow `debug` mode from any relative path - use to break as `./web` was hardcoded as static dir
- :tada: remove all `freq` calculations and replace with a much more accurate Liechtenstein distance percent. 
  - This change is pretty large and will drastically affect matches in all countries.
  - Before matches were calculated by sorting potentials with the calculated frequency of token matches. This worked well if everything was spelled correctly and helped ensure `st` vs `av` errors still lead to a match. It did however also allow things like `Margaret St` and `Shepherd St` to match on the `ST` token alone.
  - This uses lev. dist to generate a percent with a hard cutoff on if the match can take place or not. From a lot of groking the results 40% matches and above seem like a good first start.


ingalls:
cc/ @KaiBot3000 @aarthykc @sbma44 

This change once it is in should significantly affect new indexes (for the better)

ingalls:
Woot! SG went from 0.06% bad to 0.04!! Going for mergetown!!

PR #87, opened 2017-05-18
[wip] individual geocoding script, bugfixes

sbma44:
cc @ingalls

- :rocket: adds a `lib/geocode.js` that takes `lib/test.js`-like set of parameters to geocode a single query
- :rocket: modularizes geocoding capability too
- :tada: adds `--limit` param to `test.js`
- :rocket: weight linker comparisons to deemphasize mismatches based purely on tokens
    - adds `text_tokenless` column to `address_cluster` & `network_cluster` tables
    - handle all-token special case by checking for substring status
- :rocket: detection of candidate address_clusters now uses `ST_Intersects` instead of `ST_Contains`
- :bug: fix JS error in web interface related to features w/o itp data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@
 
 ## Version History
 
+### unreleased / proposed merge
+
+- :rocket: adds a `lib/geocode.js` that takes `lib/test.js`-like set of parameters to geocode a single query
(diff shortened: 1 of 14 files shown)


ingalls:
- Merged: - :bug: bugfix for zero-distance results in `test.js` mode into master

PR #90, opened 2017-05-18
Left hook

ingalls:
End of right side should be assigned closer `16`

cc/ @sbma44


ingalls:
- :white_check_mark: Add tests for hooked roads
- :rocket: Take highest/lowest number off end of road into consideration when generating ITP

PR #99, opened 2017-05-25
Create child database & allow to populate

ingalls:
For large countries with very dense network/address layers the matching time can be significant - in the order of days. 

This PR adds a new arg to the `map` module which allow for a line-delimited `(Multi)Polygon` GeoJSON file to be passed in to segment the matching by.

This allows the matching to take place in parallel on smaller regions: ie states, counties, etc.

cc/ @sbma44 @aarthykc

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -15,7 +15,7 @@ class Cluster {
      */
     address(segment = 1, cb) {
         this.pool.query(`
-            INSERT INTO address_cluster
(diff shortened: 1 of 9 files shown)


ingalls:
Segment PR moved to here: https://github.com/ingalls/pt2itp/pull/103

Leaving this PR open to eventually pull out the feature function into a separate PR.

PR #102, opened 2017-06-01
Bring back Global Tokens

aarthykc:
Passed global tokens to the Carmen instance, helped bring down the failures on Liechtenstein.

cc @ingalls

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -148,6 +148,7 @@ class Index {
 
             let tokens = tokenize.main(feat.properties.street, opts.tokens, true);
             feat.properties.street = diacritics(tokens.tokens.join(' '));       //The street is standardized and it what is used to compare to the address cluster
+            feat.properties.street = tokenize.replaceToken(tokenRegex, feat.properties.street);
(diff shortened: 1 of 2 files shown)


aarthykc:
@ingalls, could you take a look ?

lib/index.js
```diff
@@ -148,6 +148,7 @@ class Index {
 
             let tokens = tokenize.main(feat.properties.street, opts.tokens, true);
             feat.properties.street = diacritics(tokens.tokens.join(' '));       //The street is standardized and it what is used to compare to the address cluster
+            feat.properties.street = tokenize.replaceToken(tokenRegex, feat.properties.street);
```
ingalls: What exactly is this intended to do?

aarthykc: Similar to how we have it [set up in carmen](https://github.com/mapbox/carmen/pull/494/files#diff-02390562abfd478523bcb53d9efdcbb2R23) and per our conversation last week, this would be the place where we use `replaceToken` to change the regex back to text. This was there in the original global tokens PR which was wiped away with Tom's change.

ingalls: 🌮

PR #105, opened 2017-06-02
Parallel Imports

ingalls:
Will fill this out in a bit - experimenting


ingalls:
screenshot 2017-06-06 11 52 15

Satisfying Progress

PR #118, opened 2017-06-27
revise child process stdout handling to avoid colliding JSON

sbma44:
I've intermittently seen indexing fail on invalid JSON. Investigation shows what looks like `Feature`s dropped in the middle of other `Feature`s. I suspect a byproduct of the process by which numerous child workers' stdouts are piped to the same writestream.

This PR does two things to resolve the issue:
- uses the callback functionality of `process.stdout.write` to throttle writing so that it only occurs after the previous write is flushed. I don't think this helps but it's not a bad idea.
- uses the `split` module in the parent/orchestrating process to (hopefully) only write whole-line chunks to the output file

This appears to have resolved the issue for me, so I'm inclined to merge. However @ingalls is having a hard time replicating and in general this bug is a little :ghost:y.

### DIFF
--- a/lib/map.js
+++ b/lib/map.js
@@ -16,6 +16,7 @@ const Index = require('./index');
 
 const buffer = require('./buffer');
 const split = require('./split');
+const linesplit = require('split');
 const misc = require('./misc');
 
 module.exports = function(argv, cb) {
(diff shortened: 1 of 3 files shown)


ingalls:
These all look like 🚀 ideas.

ingalls:
Published as `v11.1.0`

PR #123, opened 2017-06-29
[ready] optimize adoption query

sbma44:
Adds an index, which speeds up adoption approximately 60x in my testing. Parallelization hasn't been pursued yet but could be.

I also bumped the postgres client pool size to the number of CPUs on the system -- in practice this will go unused, but was early work to support parallelization. Won't hurt to merge.

cc @ingalls


sbma44:
expanding the connection pool broke tests on circle, which apparently can't accommodate that many client connections with its current settings. I've removed that commit to keep things simple. Easy enough to keep that value down in test mode if we decide we need it later.

PR #124, opened 2017-06-29
[ready] remove low-quality network_cluster => address_cluster matches

sbma44:
This PR addresses two error conditions that are possible in master:

- low-quality linker matches where the winning `address_cluster` doesn't share a leading letter with the winning `network_cluster` (per @apendleton & @boblannon's suggestions)
- cases where a given `address_cluster` is assigned to multiple `network_cluster`s. A postprocessing step is added that reviews all such network/address groups and removes all but the best text matches.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -75,14 +75,15 @@ class Cluster {
     adoption(cb) {
         this.pool.query(`
                 BEGIN;
-		ALTER TABLE network_cluster ADD COLUMN IF NOT EXISTS address_text TEXT;
-		UPDATE network_cluster n SET address_text=a._text FROM address_cluster a WHERE n.address = a.id;
-		ALTER TABLE address_cluster ADD COLUMN orphan_adoptive_parent BIGINT;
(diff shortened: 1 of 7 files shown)


sbma44:
Tests pass locally. I will dig into what Circle's complaints are tomorrow.

sbma44:
Circle error was due to postgres 9.5 not supporting `ADD COLUMN IF NOT EXISTS`.

mattficke:
There's a funny hack to use 9.6 on Circle: https://discuss.circleci.com/t/ubuntu-14-04-build-image-update-201701-01/9245

>We only support 9.5 at this moment, but 9.6 was installed and running in the build image accidentally.
We will stop running 9.6 by default but the version will be kept installed. If you wish to use 9.6 instead of 9.5, you can add the following in your circle.yml.

sbma44:
Thanks! I think dockerification will make this less of an issue long-term. But this was just syntax sugar; for now it's easier to simply not use this feature.

Issue #128, opened 2017-07-03
text_tokenless bug


sbma44:
The highlighted `network_cluster` isn't getting matched to the adjacent `address_cluster` (around 24.707294,59.209312)

![image](https://user-images.githubusercontent.com/31717/27808124-cab1c1ce-6013-11e7-9ae7-0f3d8c7e24d2.png)

debugging in linker.js reveals that 
- the comparison between the two objects is the only one that is made -- a good start!
- the calculated distance is _exactly_ 40%, which is as high as it could be while still failing to be returned as a match

But why is it 40%? Here is the matching object:

```
{ score: 40,
  address: 
   anonymous {
     network: 'sookure',
     network_text_tokenless: 'sookure',
     id: 11964,
     text: 'sookure tanav',
     _text: 'Sookure Tänav',
     text_tokenless: 'sookure tanav' } }
```

`text_tokenless` should be just `sookure`. I've confirmed that `Tänav` is in the current version of `@mapbox/geocoder-abbreviations`. Guessing that `diacritics()` is being applied in a way that breaks the `text_tokenless` comparison but I need to confirm.

sbma44:
ok: text_tokenless appears to be broken on `address`, which leads to it being broken on `address_cluster`. Having a hard time tracking down where this all gets populated, may need to stash everything & come back to it on Wednesday w/ @ingalls around.

sbma44:
looks like this is a problem in `copy.js`, which generates PSV files that are later imported, but I haven't yet untangled how this code differentiates between `network` and `address` input.

PR #131, opened 2017-07-05
fix tokenless text in copy.js

sbma44:
refs https://github.com/ingalls/pt2itp/issues/128


sbma44:
this PR:
- adds tests for copy.js
- correctly passes tokens to `copy.js`, resolving the unset address `text_tokenless` bug
- adds a single-feature debug function for linker.js
- handles all-token features in `cluster.prune()` correctly
- moves token replacer creation code to `tokenize.js`

PR #138, opened 2017-07-10
Cardinal sequester [✈️ 🐠 Code]

ingalls:
- Add concept of post scripts
- Add cardinal switching script as first optional post script `S Main St` => `S Main St,Main  ST S`
- Add a ton of `map` test updates
- 100% coverage of minjur function
- Improve tests coverage of strip-unit.

cc/ @mapbox/geocoding-gang


ingalls:
## TODO

- ITP internal order is not stable in map.js sg test - this should not be the case as `itps.sort` is called on all output - checkout what's up. Maybe just the PT order is not stable? Very likely this.

- SG Map tests should delete id field before writing to file or else unit tests will change each time.

ingalls:
Confirmed: ITP is stable `addressnumbers` are not! 
Ref: 5e1007a66cbd8e5226f4a2d03518d5091fb19060

ingalls:
- Rewrote `id()` call to new mandatory post section to standardize all calls into a single place.

ingalls:
Tests are still fluctuating and are due to a non-deterministic part of pt2itp which I haven't been able to pinpoint as of yet. Example of changing fixture is:

One: https://gist.github.com/7112df239c8861820f89da606f49f0ba
Two: https://gist.github.com/anonymous/6eaf964c80150b98e5f0d90c12ccd560

ingalls:
Examining the diff files from the two - this is likely an issue with how the LineStrings are calculated and combined in `explode`. The `MultiPoint` clusters are exactly the same.

PR #151, opened 2017-08-10
don't attempt to geocode house numbers that are duplicates

sbma44:
A large portion of `DIST FAIL`s that I'm seeing are due to the same house number existing twice within an address cluster. I'm of two minds about this: we should definitely be aware of when the underlying data looks this way. At the same time, geocoding these is fairly pointless.

This PR adds a new overridable default behavior by which duplicate addresses are skipped and the total count is reported at the end of test mode. Some empirical results:

## before
```
------------------------------------------
DIST                          3169 (12.6%)
NAME MISMATCH                21929 (86.9%)
TEXT                           125 (0.5%)
```

## after
```
------------------------------------------
DIST                           710 (3.1%)
NAME MISMATCH                22328 (96.7%)
TEXT                            62 (0.3%)
```

cc @ingalls @aaaandrea

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -31,7 +31,8 @@ function test(argv, cb) {
                 'output',
                 'config',
                 'limit',
-                'name'
+                'name',
+                'dupes'
             ],
             alias: {
                 database: 'db',
@@ -70,6 +71,8 @@ function test(argv, cb) {
     }
 
     if (argv.limit) argv.limit = parseInt(argv.limit);
+    argv.dupes = !!argv.dupes;
+    let dupeCount = 0;
 
     const pool = new pg.Pool({
         max: 10,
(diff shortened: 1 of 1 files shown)


sbma44:
I'll add, though, that this is an alarmingly large number. The exact way that `--limit` interacts with test mode is a bit ill-defined but still...

```
# pt2itp test --config /tmp/idx.2GnHo3ub8/mapbox.places-us_nh-address-both.json --index /tmp/idx.2GnHo3ub8/dist/us_nh-address-both-64699884170810160858-64699884a1.mbtiles --db mbpl_us_nh_address_both -o ./nh-out-discardme.err --limit 2000
ok - beginning match
ok - Testing Network Matched Addresses [====================] 100% 0.0s
ok - beginning unmatch
ok - Unmatched Addresses [                    ] 0% 32.0sok - beginning diff name
ok - Name Mismatch [====================] 100% 0.0s

------------------------------------------
DIST                            15 (1.5%)
NAME MISMATCH                  992 (95.9%)
TEXT                            27 (2.6%)

ok - 1034/33580 failed to geocode
ok - skipped 20331 entries as duplicate house numbers
```

PR #152, opened 2017-08-10
optionally preserve 'source' attribute from address geojson

sbma44:
Designed to facilitate debugging of the `address` table. Useful when the input address GeoJSON is composed from several different sources.


sbma44:
cc @ingalls @aaaandrea -- def want to get your thoughts on this before I merge

aaaandrea:
I think this seems super useful. Thanks, Tom!

ingalls:
Let's get this green and merged!

Issue #154, opened 2017-08-14
mystery on Main Street


sbma44:
I'm trying to process the New Hampshire statewide data source from OpenAddresses & running into a cluster mapping bug that I can't yet diagnose.

There are a lot of unfortunately proximate MAIN STREET features in NH. These are around `-71.1244,42.9127`.

![image](https://user-images.githubusercontent.com/31717/29286040-9e5c5830-80fe-11e7-9bb0-d57081c3da9d.png)

Ignore the ones on the far right and left; we're concerned with the top and bottom. Here, I filtered for those (leaving the original though since the presence of those other nearyby features could be part of the problem, maybe). 

![image](https://user-images.githubusercontent.com/31717/29286139-eedc2038-80fe-11e7-957c-0ca63adb0ea4.png)

With our code as-is, all of these points belong to a single cluster. By commenting out the `cluster.adopt()` method, I confirmed that it's responsible for collapsing them:

![image](https://user-images.githubusercontent.com/31717/29286176-1a65e8b0-80ff-11e7-9e9a-b98cbbdd5847.png)

But why is this possible? `adopt()` should only collapse together clusters that have not been assigned to a `network_cluster`. Sure enough, that is the case here: only the top cluster belongs to a network cluster. But the bottom address cluster surrounds a correctly-named network cluster. Their `text_tokenless` fields are exact matches!

Something screwy is happening with the matching code. Next steps are to run this query on the specific IDs to see what's happening in the linker.

sbma44:
rerunning `lib/match:main` against the specific ID makes the assignment correctly. Starting to wonder if this is the product of a postgres lock problem or similar.

sbma44:
This might be at least part of the problem: nondeterministic behavior by which the bottom address cluster can get assigned to the top network segment.

![image](https://user-images.githubusercontent.com/31717/29287488-a2d8f792-8103-11e7-97ca-478c143af542.png)

I'm not positive this is the whole story but I think it's relevant and should be fixed. I'm going to adjust `match` and `linker` to be aware of situations where multiple features tie for the best match and will have the winner decided back in match.

sbma44:
It looks like https://github.com/ingalls/pt2itp/pull/156 resolves this issue

PR #156, opened 2017-08-14
don't pick match winners arbitrarily

sbma44:
`linker.js` is responsible for examining the set of address clusters that intersect a given network cluster's `buffer` box and selecting the one that has the closest matching name based on a heuristic that looks complex but is mostly about string distance (subject to a minimum match quality threshold). 

`linker` presently only returns zero or one matching features, but it is possible for the set of input addresses to contain multiple features that meet the matching heuristic identically well. In those circumstances the winner is determined by the vagaries of postgresql row ordering -- it's basically random. This leads to circumstances like https://github.com/ingalls/pt2itp/issues/154

This PR adjusts `linker`'s behavior to return all candidate addresses that score maximally on the matching heuristic. When this produces a single matching address, behavior is unchanged.

When more than one address is returned, an additional query is performed to determine which matching address cluster contains the most points falling within the network cluster's `buffer` box; this tiebreak mechanism determines the address cluster that is assigned to the network cluster.

cc @ingalls @aaaandrea

### DIFF
--- a/lib/linker.js
+++ b/lib/linker.js
@@ -5,13 +5,14 @@ const diacritics = require('diacritics').remove
  * Generates a map of street => address
  */
 module.exports = (street, addresses) => {
-    let current;
+    let maxScore = false;
 
     for (let addr_it = 0; addr_it < addresses.length; addr_it++) {
         let address = addresses[addr_it];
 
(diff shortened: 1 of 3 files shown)


ingalls:
```
📦 v11.14.0
```

PR #160, opened 2017-08-15
optimize test mode

sbma44:
dupe-avoidance code currently looks for ungeocodable duplicates: entries in address clusters with the same house number that are too far apart for there to be hope of reliably getting a correct answer.

This is still inefficient, though: the duplicates that are close together are getting geocoded multiple times. The indeterminacy about which point comes back is still present, it's just that all of the possible points are close enough together that any of them will count as correct. This circumstance should be expected any time there are overlapping data sources (which will be often).

This PR addresses this inefficiency by only geocoding the first of a duplicate, close-together set of points. The number of points skipped is tracked, as is the number of ungeocodable points. The two are displayed separately at the end of a test run.

cc @ingalls @aaaandrea

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -72,7 +72,7 @@ function test(argv, cb) {
 
     if (argv.limit) argv.limit = parseInt(argv.limit);
     argv.dupes = !!argv.dupes;
-    let dupeCount = 0;
+    let dupeCount = { ungeocodable: 0, duplicative: 0 };
 
     const pool = new pg.Pool({
         max: 10,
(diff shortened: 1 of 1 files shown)


sbma44:
sample output:

```
...
ok - Unmatched Addresses [====================] 100% 0.0s
ok - beginning diff name
ok - Name Mismatch [====================] 100% 0.0s

------------------------------------------
DIST                            81 (0.4%)
NAME MISMATCH                21957 (99.3%)
TEXT                            64 (0.3%)

ok - 22102/439358 failed to geocode
ok - skipped 273486 duplicative & proximate house numbers as redundant (these are fine)
ok - skipped 3910 duplicative & distant house numbers as impossible to geocode (these are less fine)
```

Issue #146, opened 2017-08-02
Titlecase & Cardinals


ingalls:
Input: `3134 D ST NE`

Current Incorrect Output: `3134 D St Ne`
Expected Output: `3134 D St NE`

Also Applies to
```
Northeast => NE
Northwest => NW
Southeast => SE
Southwest => SW
```

cc/ @sbma44

sbma44:
on it

ingalls:
Tracking Ref: https://github.com/ingalls/pt2itp/pull/161

PR #161, opened 2017-08-15
[ready] handle english cardinality

sbma44:
refs https://github.com/ingalls/pt2itp/issues/146

todo:
- confirm mixed case behavior in tests are desirable
- update package-lock
- bump version, publish

cc @ingalls @aaaandrea


sbma44:
thinking about this more, I don't think we want to handle `N.E` or `NE.` -- these should be titlecased as normal. Updated the dep to only handle `N.E.` or `NE`.

sbma44:
Also decided to standardize on period-less cardinals. Consistency is desirable and I believe it will minimize token count (and distinctiveness) in carmen.

PR #163, opened 2017-08-18
Detect Duplicates

ingalls:
- Detect Duplicate `> 1km` in distance apart.
- If no duplicates are detected fallback to current `explode#split` module
- if duplicates are detected pass to to-be-written `cluster#split` module

cc/ @sbma44 @aaaandrea


lib/split.js
```diff
@@ -148,7 +148,19 @@ function split(nid, cb) {
             features: [ turf.feature(network) ]
         });
 
-        network = explode.split(network, coords, number);
+
+
+        //Detect cases where there are duplicate addresses outside of the threshold
+        for (let num_it = 0; num_it < number.length; num_it++) {
+            if (!number[num_it + 1]) break; //End of the road
```
sbma44: can we just have the for loop go to .length - 1? a bit simpler

PR #164, opened 2017-08-21
Add Generic ASCII=>CIVIC Encoder Decoder

ingalls:
The current UnitMap Implementation works great as long as it is running on fewer than 10 cores.

In the current process the UnitMap number is prepended with the process #. Once this passes 10 you can get collisions between process numbers.

This PR fixes this by implementing a new ASCII based lookup encoder/decoder which will also reduce UnitMap size as all processes will have deterministic unit numbers.


ingalls:
This should be ready to go! Going to test DC to ensure unit numbers come out correctly.

PR #171, opened 2017-08-31
Explode Module can output ~0 length linestrings

ingalls:
```
{"type":"Feature","properties":{},"geometry":{"type":"LineString","coordinates":[[-83.972619,36.82752099999999],[-83.972619,36.827521]]}}
```

Is an example of a ~0 length LineString that chokes in the interpolate module. These shouldn't be created. First pass is to see if this is coming from Explode#Split. If this still errors will add specific handling in interpolate.

cc/ @sbma44 @aaaandrea


ingalls:
Despite this fix still seeing a null length geom enter interpolate.js. Going to also add this check to the interpolate side of things.

PR #173, opened 2017-08-31
Use CSV Library for PSV creation in copy

ingalls:
We're dramatically increasing the number of data sources we're testing in PT2ITP and it's only a matter of time before we hit a bug with the current homebrewed PSV generator.

This PR implements the use of a CSV library for ensuring proper escaping and PSV creation.

cc/ @sbma44 @aaaandrea


sbma44:
👏

Issue #172, opened 2017-08-31
optimized adoption code breaks singleton address clusters


sbma44:
@ingalls and I figured this one out just before I had to leave today: https://github.com/ingalls/pt2itp/pull/170 removes all uniquely-named address_clusters from consideration in the adoption code. But _every_ cluster has its geometry recombined. This change makes it impossible for singletons to access themselves, and they wind up with empty geometry.

The fix should be to make the `UPDATE` not affect all clusters. This is likely to be a bigger optimization than what we merged, anyway.

sbma44:
fixed in #174

PR #174, opened 2017-08-31
optimize adoption query, squash bug, add tests

sbma44:
refs https://github.com/ingalls/pt2itp/issues/172

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -226,7 +226,7 @@ class Cluster {
                  SET name_uniqueness = a.cnt
                  FROM (SELECT COUNT(id) AS cnt FROM address_cluster GROUP BY _text) a;
 
-            ALTER TABLE address_cluster ADD COLUMN orphan_adoptive_parent BIGINT;
+            ALTER TABLE address_cluster ADD COLUMN orphan_adoptive_parent BIGINT DEFAULT NULL;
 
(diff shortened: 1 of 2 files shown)


sbma44:
looks good in WV test case. prior to this fix you could see `address` points peaking out from underneath the `address_cluster` layer

image

sbma44:
this appears to have made things quite a bit slower -- need to dig into the query a bit more & pursue optimizations.

sbma44:
broke this up & timed it with a small patch of WV. I'm not convinced this is _totally_ representative because of the low number of updated features for the later tasks. But I think the name_uniqueness query definitely needs some work asap.

| task | duration (s) |
|------|--------------|
| setting address_text | 0.34 |
| setting name_uniqueness | 28.99 |
| setting adoptive parent | 0.15 |
| creating orphan groups | 0.07 |
| updating geom | 0.05 |
| deleting spares | 0.05 |

sbma44:
OK, big changes:

- move `_text` index creation to top of block so it's available at the time of the name_uniqueness setup
- create `name_count` as temp table instead of subquery
- create an index on network_cluster.address

| task | duration (s) |
|------|--------------|
| setting address_text | 0.34 |
| setting name_uniqueness | 0.24 |
| setting adoptive parent | 0.55 |
| creating orphan groups | 0.09 |
| updating geom | 0.09 |
| deleting spares | 0.08 |

PR #177, opened 2017-09-07
Fixing result logging for "NO RESULTS"

boblannon:
This is where the mysterious "N" result code was coming from.  While all of the other results were arrays, this one was a string.  So in `addrQ.awaitAll` below:

```javascript
                                 for (let r of res) {
                                     if (r === true) continue;
                                     logResult(r[0], r[1]);
                                 }
```

 `r[0]` is called on the string (which starts with "N"), rather than on an array, as it does for the other result types.

@sbma44 one beer please.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -208,13 +208,14 @@ function test(argv, cb) {
                                         c.geocode(query, opts, (err, res) => {
                                             if (err) return done(err.toString());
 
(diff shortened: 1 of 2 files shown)


sbma44:
HERO! Whyyy could I not find this myself. Ugh. Thank you. What kind of beer are you into these days?

boblannon:
terrapin's T-Time, but unfortunately it turned out to be seasonal. surprise me! 🍻

Issue #180, opened 2017-09-07
[proposal] supporting multiple names for network clusters


sbma44:
consider this real-world scenario:

![image](https://user-images.githubusercontent.com/31717/30174151-17b726a2-93c8-11e7-8661-a470da42de3c.png)

The network line in question has both names assigned to it in our source data -- this assignment can be made... theoretically.

So what if we just added multiple names to the network_cluster and ran each one through `linker`? It'd work fine for this:

![image](https://user-images.githubusercontent.com/31717/30180393-bd167804-93dd-11e7-8f3a-23693419f160.png)

What do we do with the network_cluster when the alternate name(s) only exist on part of the road? This is pretty common. Breaking it into three individual `network_cluster`s is an obvious strategy. And that would work sometimes, like here:
![image](https://user-images.githubusercontent.com/31717/30180485-0a7d0bd0-93de-11e7-9ce4-c5eb68fe5dfd.png)

But we have to consider cases when the points are labeled in a manner that doesn't totally correspond to where the network data says it starts and ends. This is also pretty common:
![image](https://user-images.githubusercontent.com/31717/30180660-9dd812f8-93de-11e7-90be-b4583b4b9fe5.png)
![image](https://user-images.githubusercontent.com/31717/30180675-ac3c6844-93de-11e7-99fe-6319507b2297.png)

The bottom case is especially bad, because the address cluster would be assigned to one of the `Rt 10` segments but not all three, producing obviously bad interpolation output.

----------

So what do we do? I think at ingestion we should add a new `network` table row for each synonym attached to a road segment. Clustering proceeds as normal, producing overlapping road geometry that can be assigned to whichever address cluster matches its name.

What happens in the third case above? Well, two options. In the simplest implementation, we just get bad interpolation data for the central segment.

But! I think it would be preferable to store original network segment IDs (e.g. OSM IDs) in the `network` table and then pass these through to `network_cluster` with an `ARRAY_AGG(source_id) AS source_ids`. We then select groups of `network_cluster`s where `a.source_idd && b.source_idd AND ST_Intersects(a.buffer, b.buffer)` and consider whether to merge them. Here's some pseudocode that I am pretty sure will contain bugs:

```
output_segments = [ longest(network_segments) ]
remaining_segments = network_segments - longest(network_segments)
for seg in remaining_segments:

  # look for coextensive segments & collapse together
  merged_with_twin = false
  for o in output_segments:
    if o.source_ids == seg.source_ids:
[...]

PR #182, opened 2017-09-08
[ready] network multilabel

sbma44:
refs https://github.com/ingalls/pt2itp/issues/180


test/cluster.clusterWithin.test.js
```diff
@@ -134,8 +134,8 @@ test('LinesStrings far away should not be clustered', (t) => {
         pool.query(`
             BEGIN;
             DROP TABLE IF EXISTS network;
-            CREATE TABLE network (id SERIAL, segment BIGINT, text TEXT, text_tokenless TEXT, _text TEXT, named BOOLEAN, geom GEOMETRY(LINESTRING, 4326));
-            CREATE TABLE network_cluster(ID SERIAL, text TEXT, _text TEXT, text_tokenless TEXT, geom GEOMETRY(GEOMETRY, 4326), buffer GEOMETRY(Polygon,4326), address INTEGER);
+            CREATE TABLE network (id SERIAL, segment BIGINT, text TEXT, text_tokenless TEXT, _text TEXT, named BOOLEAN, geom GEOMETRY(LINESTRINGZ, 4326));
```
boblannon: the type for `id` should be changed to `BIGINT`, to match what's going on in `copy.js` now

boblannon:
got this to run on US MT.  here are the side-by-side numbers before and after the change:

```
| ERROR TYPE             | BEFORE | AFTER  |
|------------------------|--------|--------|
| DIST                   | 21     | 22     |
| NAME MISMATCH (HARD)   | 864    | 2342   |
| NAME MISMATCH (SOFT)   | 404    | 374    |
| NOT MATCHED TO NETWORK | 108075 | 101532 |
| TEXT                   | 259    | 137    |
```

I'll dig more into those tomorrow but `NOT MATCHED TO NETWORK` has gone down by thousands.  `NAME MISMATCH (HARD)` seems to have increased, though, so i'll check into that.

sbma44:
Hmm. I was hoping for a much bigger improvement than this.

boblannon:
yeah, me too.  my back-of-the-envelope estimation was around 38k addresses that had something like "highway" in their names.

boblannon:
the endcap param for ST_Buffer might be useful when trying to find overlap between two conterminous linestrings: 

2017-09-29 at 10 33 am

sbma44:
@boblannon ah but I'm preserving the unique IDs of each geometry fragment -- no geometric operations are necessary to determine overlap, just some set operations.

sbma44:
(but I should add that @ingalls does use techniques just like this in other parts of the codebase when relating network clusters to address clusters)

sbma44:
OK! This is ready for a review & merge. **BUT FIRST WE SHOULD CONFIRM THAT IT ACTUALLY HELPS.** I feel good about the unit tests in place; I feel good about the need for this code; I feel good about the implementation. But the proof is in the passrate.

@ingalls, would love a review.

ingalls:
I've tested this out on several different sources now and not seeing any alarms. PT2ITP test mode shows imporved passrates. Going to go forward with a merge here. Thanks a ton for all your hard work @sbma44!

ingalls:
:taco: :tada:

PR #184, opened 2017-09-12
changing any null text fields to empty string

boblannon:
Null values in `address` and `network` tables' text fields shouldn't be allowed, but they're getting created because the `COPY` commands are treating empty strings as null:

```javascript
for (let child of nursery) {
    netQ.defer((output, done) => {
        self.pool.query(`
            COPY network (text, text_tokenless, _text, geomtext)
            FROM '${output}' WITH CSV DELIMITER '|' QUOTE E'\b' NULL AS '';
        `, (err) => {
            if (err) return done(err);

            fs.unlink(output, (err) => {
                return done(err);
            });
        });
    }, child.output);
}
```

That's fine, we need it for other values that are stored as empty strings in the PSV file.  it has the unfortunate side effect, though, of creating null values in the text fields.

```
mbpl_us_mt_address_both=# select count(*) from network;
 count
--------
 105584
(1 row)

mbpl_us_mt_address_both=# select count(*) from network where _text is null;
 count
-------
 45222
(1 row)
```

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -417,27 +417,41 @@ class Cluster {
 
                 let str = Object.keys(names);
 
-                str.sort((a, b) => {
-                    return names[b] - names[a];
-                });
 
-                if (!str.length) return cb();
+                if (!str.length) {
+                    this.pool.query(`
+                        UPDATE
+                            network
+                        SET
+                            named = FALSE
+                        WHERE
+                            id = ${id}
+                    `, (err, res) => {
+                        return cb(err);
+                    });
+                }
+                else {
+                    str.sort((a, b) => {
+                        return names[b] - names[a];
+                    });
 
-                let text = JSON.parse(str[0]);
+                    let text = JSON.parse(str[0]);
 
-                this.pool.query(`
-                    UPDATE
-                        network
-                    SET
-                        text = '${text[0]}',
-                        text_tokenless = '${text[1]}',
-                        _text = '${text[2]}',
-                        named = TRUE
-                    WHERE
-                        id = ${id}
-                `, (err, res) => {
-                    return cb(err);
-                });
+
+                    this.pool.query(`
+                        UPDATE
+                            network
+                        SET
+                            text = '${text[0]}',
+                            text_tokenless = '${text[1]}',
+                            _text = '${text[2]}',
+                            named = TRUE
+                        WHERE
+                            id = ${id}
+                    `, (err, res) => {
+                        return cb(err);
+                    });
+                }
             });
         });
     }
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -418,7 +418,7 @@ class Cluster {
                 let str = Object.keys(names);
 
 
-                if (!str.length) {
+                if (res.rows.length == 0 || !str.length) {
                     this.pool.query(`
                         UPDATE
                             network
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -408,20 +408,23 @@ class Cluster {
                 if (err) return cb(err);
 
                 for (let res_it = 0; res_it < res.rows.length; res_it++) {
-                    let text = JSON.stringify([
-                        res.rows[res_it].text,
-                        res.rows[res_it].text_tokenless,
-                        res.rows[res_it]._text
-                    ]);
+                    // only consider results that have non-empty _text strings
+                    if (res.rows[res_it]._text) {
+                        let text = JSON.stringify([
+                            res.rows[res_it].text,
+                            res.rows[res_it].text_tokenless,
+                            res.rows[res_it]._text
+                        ]);
 
-                    if (!names[text]) names[text] = 1;
-                    else names[text]++;
+                        if (!names[text]) names[text] = 1;
+                        else names[text]++;
+                    }
                 }
 
-                let str = Object.keys(names);
+                let namelist = Object.keys(names);
 
 
-                if (res.rows.length == 0 || !str.length) {
+                if (res.rows.length == 0 || !namelist.length) {
                     this.pool.query(`
                         UPDATE
                             network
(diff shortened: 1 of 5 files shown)


boblannon:
this runs into trouble in `cluster.name`, which is a function that may well have never been called on real data before. that is, if the conditions for calling it involved empty-string values in network text fields, then those conditions might never have been met before these changes.

here's the stack trace i'm getting:  
```
TypeError: Cannot read property 'pool' of null
    at name (/usr/local/src/pt2itp/lib/cluster.js:395:13)
    at start (/usr/local/src/pt2itp/node_modules/d3-queue/build/d3-queue.js:74:11)
    at poke (/usr/local/src/pt2itp/node_modules/d3-queue/build/d3-queue.js:58:11)
    at Queue.defer (/usr/local/src/pt2itp/node_modules/d3-queue/build/d3-queue.js:33:5)
    at cursor.read (/usr/local/src/pt2itp/lib/map.js:215:35)
    at Immediate.setImmediate (/usr/local/src/pt2itp/node_modules/pg-cursor/index.js:86:7)
    at runCallback (timers.js:666:20)
    at tryOnImmediate (timers.js:639:5)
    at processImmediate [as _immediateCallback] (timers.js:611:5)
```

...which is a little baffling, because i'm not sure how `this` could evaluate to null.  I'm going to continue looking into it, though.

boblannon:
Okay, so I'm documenting my debugging process in case it's helpful for others.  To get an example of a thing that was failing, I added a `try/catch` to the `name` function in `lib/cluster.js`:

```javascript
    name(id, cb) {
        try{
            // content of the function
        }
        catch (err){
            console.error(`name failed for ${id}`);
            return cb(err);
        }
    }
```

This gave me some extra output before the error message above:

```
name failed for 90211
TypeError: Cannot read property 'pool' of null
    at name (/usr/local/src/pt2itp/lib/cluster.js:396:13)
```

So now I know that the `address` record with `id=90211` is the problematic case. I grabbed its `geom` field, which is a LineString:

```sql
select
    text,
    _text,
    text_tokenless,
    ST_AsGeoJSON(geom)
from
    network
where
    id = 90211;
```

(turned on extended display in psql using `\x on`, which is helpful for when some fields are really long):

```
-[ RECORD 1 ]--+---------------
text           |
_text          |
text_tokenless |
st_asgeojson   | {"type":"LineString","coordinates":[[-108.825427,35.458227],[-108.825914,35.458147]]}
```

So definitely a candidate for the `name` function, since its text fields are empty strings. My guess was that it wasn't finding any matches in the address table. In the code, the matches are found using a custom buffer function [buffer.js](https://github.com/ingalls/pt2itp/blob/master/lib/buffer.js):

```javascript
let str = JSON.parse(res.rows[0].geom);
let buff = buffer(str, 0.1).geometry

this.pool.query(`SELECT text, text_tokenless, _text FROM address WHERE ST_Within(geom, ST_SetSRID(ST_GeomFromGeoJSON('${JSON.stringify(buff)}'), 4326))`, (err, res) => {
   // do stuff
})
```

...where `res.rows[0].geom` is the same value I just got from the network query above.

I wanted to get an idea of how this all looked. I opened a node interpreter to manually re-create the poly that `buffer` would return.

```
> const buffer = require('./lib/buffer')
> d =  {"type":"LineString","coordinates":[[-108.825427,35.458227],[-108.825914,35.458147]]}
> buff = buffer(d, 0.1)
> JSON.stringify(buff)
'{"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-108.82520880286363,35.45734570351165],[-108.82569579774577,35.45726570438799],[-108.8261322070349,35.45902829521931],[-108.82564520191697,35.4591082960957],[-108.82520880286363,35.45734570351165]]]}}'
```

Then I used the postgis function [ST_Buffer](http://www.postgis.net/docs/ST_Buffer.html) to find the nearest matches to that record's `geom`. NB: this doesn't behave exactly like pt2itp's buffer function. In particular, the numeric argument seems much more sensitive in `ST_buffer`. 

```sql
select
    ST_AsGeoJSON(ST_Union(a.geom)),
    a.text,
    a.text_tokenless,
    a._text
from
    address a
        join
    network n on ST_Within(a.geom, ST_Buffer(n.geom,0.01))
where
    n.id = 90211
group by a.text, a.text_tokenless, a._text;
```


```
-[ RECORD 1 ]--+------
st_asgeojson   | {"type":"MultiPoint","coordinates":[[-108.82868,35.458087,336],[-108.82654,35.464779,227],[-108.82615,35.459136,349],[-108.82608,35.457165,356],[-108.82485,35.449949,424],[-108.82153,35.456137,381],[-108.82071,35.454032,389]]}
text           | twin buttes rd rd
text_tokenless | twin buttes
_text          | TWIN BUTTES ROAD RD
```

So now I have enough data to make three features:
 - The LineString of the original network record
 - The Polygon of the buffer created by `buffer.js`
 - The Multipoint collection of points that are nearest to the LineString

I put all of those parts together (semi-manually) in geojson.io ([gist here](https://gist.github.com/boblannon/89486a7af87e24cf40515384ae8a706e)). Turns out that yes, it looks like this was a network record that couldn't find any nearby matches:

2017-09-13 at 1 58 pm

boblannon:
Open questions:
- why doesn't the new check catch the cases where nothing was matched?
- why not just use postgis `ST_Buffer`?

Possible solutions:
- better exception handling to fail more gracefully in the case of true (and unfixable) misses
    - sortof seems like this needs to happen anyway
- make the radius of the buffer slightly larger
    - might be hard to tell what effects this would have
    - any way to optimize the radius parameter based on ground truth data?
- see if `ST_Buffer` does any better
    - there's at least one potential advantage, which is that you can (optionally) have endcaps.  that might have avoided this miss, for instance:
       2017-09-13 at 2 58 pm

boblannon:
It's possible that this code is working now, but if so, it does seem to take an awfully long time.

ingalls:
:taco: looks good to me

PR #188, opened 2017-09-20
Missed closing brackets

aarthykc:
Currently, we're running into:
```
node lib/test.js
pt2itp/lib/test.js:400
                        });
                         ^
SyntaxError: Unexpected token )
```
We're missing a closing bracket on the line before it, so this PR adds the closing bracket. Tested it out locally seems to work ! 

cc @ingalls


ingalls:
:package:` v11.22.2`

PR #198, opened 2017-10-04
Philosophy of Ephemerality

ingalls:
Often OpenAddresses won't have complete data for a county, resulting in subpar ITP results for that county. Where possible we find & improve the OA source data but this isn't always possible.

This PR introduces a second strategy which allows one to import ephemeral points which can be used to build the ITP network but then are dropped before they are output. (Not included in the final address points output)

@iandees, I know at one point you were talking or had written tiger interpolation => pt scripts. Did ever end up completing them? If not I have the scripts which I would be happy to share.


iandees:
I didn't write the code. It'd be interesting to see it!

ingalls:
@iandees Eventual plan is to clean this up into a nice CLI utility but for the time being it's here: https://gist.github.com/ingalls/69054e8d920958f4e48e00e8b561c4f7

PR #205, opened 2017-10-10
Include "ref" string as a synonym for OSM network features

boblannon:
Per #182, in addition to `name`, we are using `loc_name` and `alt_name` as synonyms to increase the surface area for address to network matches.  In some cases, though, [the `ref` property](https://wiki.openstreetmap.org/wiki/Key:ref) is also useful.

Here's a (simplified) example, from [Sylvania Road](https://www.openstreetmap.org/way/458585568#map=18/41.80409/-76.87708), which is a segment of US Route 6.

```xml
<way id="458585568" visible="true" version="2" changeset="44637042" timestamp="2016-12-24T01:22:06Z" user="Albert Pundt" uid="734072">
    <nd ref="4545687181"/>
    <tag k="alt_name" v="Grand Army of the Republic Highway"/>
    <tag k="name" v="Sylvania Road"/>
    <tag k="ref" v="US 6"/>
    <tag k="ref:penndot" v="SR 0006"/>
</way>
```

This `way` object is a member of the US 6 relations ([east](https://www.openstreetmap.org/relation/6826140) and [west](https://www.openstreetmap.org/relation/6826139)), and ultimately their parent relation [US 6 (super)](https://www.openstreetmap.org/relation/443222).

There are two options for improving our matches:

1. Make use of the `ref` field, which seems to percolate down the `name` value of the parent relations
2. Actually recover the name "US 6" from the parent relations. 

Option (2) would involve some complicated relation tree climbing, and would only be worth it if the `ref` field is unreliable for some reason.

So this PR goes with option (1).  It's a small change to the code, but it shouldn't be merged until I can look at the impact on the results.


boblannon:
Tested these changes on PA, SC, NV and MT.  They don't have a huge impact, but some, and it's positive.

PR #207, opened 2017-10-12
remove record separator during `convert`

boblannon:
This is a small change that strips the control character [record separator](http://www.fileformat.info/info/unicode/char/001e/index.htm) from input lines to avoid a `JSON.parse` error.


boblannon:
@Ingalls says " 🌮 " IRL

PR #211, opened 2017-10-23
Externalize Test Mode

ingalls:
Add support for testing External CSVs and modularize some of the current hardcoded logic in `test` mode.


ingalls:
Sample output for OpenAddresses US/DC data comparing to known business addresses

```
DIST                            24 ( 35.8% of errors |  0.8% of total addresses)
NO RESULTS                      15 ( 22.4% of errors |  0.5% of total addresses)
TEXT                            28 ( 41.8% of errors |  1.0% of total addresses)

ok - 67/2938 (2.3%) failed to geocode
```

ingalls:
This morphed into a bigger change than first intended

- Migrate to Circle2.0
- Remove native `copy` from `map` mode. This is due to the fact that `copy` assumes that the data is on the same machine as the caller. This is potentially not the case with Docker/Circle where the caller can be on a different machine calling a remote postgres function. Replace this with copying via a raw input stream and a ` COPY FROM STDIN`.

PR #212, opened 2017-10-24
Handle Octothorps in Highways like `HWY #35`

KaiBot3000:
Fixes odd highway formatting that was keeping addresses from matching. 

- [x] add `#` stripping regex function `reformatUSHighway`
- [x] add testcase


ingalls:
:taco:

PR #215, opened 2017-10-26
Addresses as Free Radicals

ingalls:
> Why is it that when robots are stored in an
> empty space, they will group together,
> rather than stand alone? How do we explain
> this behavior? Random segments of code?
>Or is it something more?

The Current Behavior Visualized

```
------N1------|-------N2-------
..............A1...............
......A2......
              ........A3.......

N1=Main St W
N2=Main St E
A1=Main St
A2=Main St W
A3=Main St E

N1=A2 Since it shares name + cardinal
N2=A3 ditto
A1 is left orphaned or joined to a random street
    it kinda matches b/c it has a diferent name
    than A2/A3
```

The problem is PT2ITP doesn't attempt to
collapse similiary named address clusters
together. It assumes that each name is a
unique street. This assumption needs to be
updated to instead cluster all of the nets
together and then match each addr with its
geometric & textually proximal net. This
would result in A2 & A3 getting matched
with N1 and N2 as currently and then A1
would be split into 2 different groups,
half with N1 and half with N2.

In addition to the split it will negate
the need for adoption code as well as
significantly improve our handling of unit 
numbers accidently retained in the street
name. As currently a unique address
cluster is created for each of these.


ingalls:
- [ ] `adoption` is slated for removal
- [ ] `cluster.address` will need to be retained but will only be run on addresses that are not successfully matched to  a network segment and will augment the current `orphan.address `
- [ ] Add unit tests for above issue
- [ ] How are synonyms going to work now.. These are currently added as identical streets (same geom) probs don't need to exactly do this now that we aren't confined to 1:1

ingalls:
the current design philosophy for this PR was to deprecate clusters on the postgres side with the exception of the orphan address clustering which @KaiBot3000 is implementing.

Upon further review this appears to be wildly problematic as it would require a rewrite of the `cluster#collapse` functions which are based on merging ids on the network together. Since these ids are inverse now ie `address => network` instead of `network => address` we would take another perf hit in addition to the perf hit we will take in the new match behavior.

To avoid this the `address` table will be clustered using `ST_Collect` after the matching operation has completed on addr features that share the same network id. The `network.address` property will be reimplemented and backfilled from the new cluster behavior.

sbma44:
possibly OT but: any concerns about performance when addresses aren't clustered? I understood that to b the reason for the original architecture...

lib/match.js
```diff
@@ -48,74 +41,126 @@ function kill() {
...
         FROM
-            address_cluster addr,
-            network_cluster AS network
+            address a
         WHERE
-            network.id = ${id} AND
-            ST_Intersects(network.buffer, addr.geom);
+            a.id >= ${min} AND a.id <= ${max}
+        GROUP BY
+            a.id;
```
sbma44: is a.id not unique?

lib/orphan.js
```diff
@@ -24,79 +25,88 @@ class Orphan {
...
-                                    coordinates: []
-                                }]
+        const cluster = new Cluster(this.pool);
+
+        cluster.orphan((err) => {
+
+            this.pool.connect((err, client, done) => {
+                if (err) return cb(err);
+
+                const cursor = client.query(new Cursor(`
```
sbma44: when is address_orphan_cluster actually populated?

KaiBot3000: https://github.com/ingalls/pt2itp/blob/FreeRadicals/lib/cluster.js#L155

lib/match.js
ingalls: @sbma44 because of the way the subquery is being processed it was very unhappy with me not grouping. a.id is a `SERIAL` so definitely unique.

ingalls:
Addresses being incorrectly matched in DC

| Address | Note |
| --- | --- |
| ` n st nw` | @KaiBot3000 is working on this bug |
| `r st nw` | @KaiBot3000 |

ingalls:
screenshot from 2017-11-13 11-23-50

PR #216, opened 2017-10-30
Add Test Case for Free Radicals (address clustering -> solo addresses)

KaiBot3000:
Builds test for #215 

- [x] creates sample `address` and `network` geojson with slight cardinality differences that `FreeRadicals` should fix
- [x] adds stub test that runs ^ through `map.js`
- [ ] actually test for something

The first pass at a test had a couple problems; the test would either hang or error out, apparently at random. 

- [x] tidy up how we track child processes, to make sure they're being closed properly
    - [x] don't increment the `length` property of an array
    - [x] check the `killed` property of a child process directly rather than relying on a separately `active` property

- [x] don't write empty lines from `split` child processes, which was causing a `write after end` error

We worked through the errors, but the test is still failing to end as expected.

### DIFF
--- a/lib/map.js
+++ b/lib/map.js
@@ -170,7 +170,7 @@ function main(argv, cb) {
     }
 
     /**
-     * Attempt to add names to any unanmed streets
+     * Attempt to add names to any unnamed streets
      * @param {Error} err Any errors from parent function/step
      */
     function nameNetwork(err) {
@@ -348,7 +348,7 @@ function main(argv, cb) {
      * . . .                .   .   .
      *   A                      B
      *
-     * By default Line 1 can only match with A or B not both,  detect these cases and group A & B together into a single cluster
+     * By default Line 1 can only match with A or B not both, detect these cases and group A & B together into a single cluster
      */
     function adopt() {
         return cluster.adoption(() => {
@@ -372,7 +372,7 @@ function main(argv, cb) {
     }
 
     /**
-     * Find clustest Segment in network for each address point in matched cluster and generate ITP output
+     * Find closest Segment in network for each address point in matched cluster and generate ITP output
      */
     function splitter() {
         pool.connect((err, client, pg_done) => {
--- a/lib/map.js
+++ b/lib/map.js
@@ -404,7 +404,8 @@ function main(argv, cb) {
                     child.stdout
                         .pipe(linesplit())
                         .on('data', (line) => {
-                            output.write(line + '\n');
+                            console.log('\n ok - splitter line:', line);
+                            if (line) output.write(line + '\n');
                         });
                     child.stderr.pipe(process.stderr);
 
@@ -417,7 +418,7 @@ function main(argv, cb) {
                             if (message.type === 'end') {
                                 // type == end means we have received the all-clear from a process that acted on a kill signal
                                 // nursery[message.id].active = false;
-                                console.log('\nkiddos:', nursery.length, nursery[message.id].killed);
+                                console.log('\nkiddos:', nursery.length, ", killed:", nursery[message.id].killed);
                                 nursery[message.id].kill();
 
                                 let active = nursery.some((instance) => {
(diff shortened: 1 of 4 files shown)


KaiBot3000:
The test is still hanging, and I'm sure all the child processes are successfully ended: 

### during a test run (note running child process):

image


### during the hang after the test should be finished:
image


still poking.

KaiBot3000:
I knew from ^ it wasn't a child process, so I looked for open psql connections:
```
SELECT datname,usename,pid,client_addr,wait_event,query_start,query FROM pg_stat_activity;
```
and found the open `query` was in [`cluster.prune`](https://github.com/ingalls/pt2itp/blob/master/lib/cluster.js#L346):
```
 datname  | usename  |  pid  | client_addr | wait_event |          query_start          |                                                             query                                                              
----------+----------+-------+-------------+------------+-------------------------------+--------------------------------------------------------------------------------------------------------------------------------
 postgres | postgres | 33244 |             |            | 2017-10-30 11:04:32.097978-07 | SELECT datname,usename,pid,client_addr,wait_event,query_start,query FROM pg_stat_activity;
 pt_test  | postgres | 32930 | 127.0.0.1   |            | 2017-10-30 11:01:03.86988-07  |                                                                                                                               +
          |          |       |             |            |                               |                 SELECT                                                                                                        +
          |          |       |             |            |                               |                     a.id AS address_id,                                                                                       +
          |          |       |             |            |                               |                     a.text AS address_text,                                                                                   +
          |          |       |             |            |                               |                     a.text_tokenless AS address_text_tokenless,                                                               +
          |          |       |             |            |                               |                     n.id AS network_id,                                                                                       +
          |          |       |             |            |                               |                     n.text AS network_text,                                                                                   +
          |          |       |             |            |                               |                     n.text_tokenless AS network_text_tokenless                                                                +
          |          |       |             |            |                               |                 FROM network_cluster n inner join address_cluster a ON n.address=a.id                                         +
          |          |       |             |            |                               |                 WHERE n.address IN                                                                                            +
          |          |       |             |            |                               |                     (SELECT address FROM network_cluster WHERE address IS NOT NULL GROUP BY address HAVING COUNT(address) > 1)+
          |          |       |             |            |                               |                 ORDER BY a.id ASC; 
```
Turns out the `pg_done` [here](https://github.com/ingalls/pt2itp/blob/master/lib/cluster.js#L398) was never called because there was an err triggering the callback:
```javascript
                            that.pool.query(`UPDATE network_cluster SET address=NULL WHERE id IN (${deleteMe});`, (err, res) => {
                                if (err) return cb(err);
                                pg_done(); // never called, leaving connections open
                                return cb();
                            });
```

 I console'd out the error:
```
{ error: syntax error at or near ")"
    at Connection.parseE (/Users/kai/src/pt2itp/node_modules/pg/lib/connection.js:546:11)
    at Connection.parseMessage (/Users/kai/src/pt2itp/node_modules/pg/lib/connection.js:371:19)
    at Socket.<anonymous> (/Users/kai/src/pt2itp/node_modules/pg/lib/connection.js:114:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:551:20)
```
but because of the way we're piping into postgres this will be very difficult to track down. @ingalls suggested moving `pg_done` above the error handling, and that keeps the test from hanging.

🎉 🎉 🎉 

Next Steps:
- [ ] clean up debugging code
- [ ] move `pg_done` above error handling *everywhere* (at nick's suggestion) to keep us from falling in this hole again

KaiBot3000:
I couldn't leave it :P 
The syntax error was coming from `UPDATE network_cluster SET address=NULL WHERE id IN (${deleteMe});`, which is invalid when `deleteMe === []`. I added `deleteMe = deleteMe ? deleteMe : [-1];` to keep the SQL valid even if we're keeping all of our networks. Note that this assumes (and eventually, guarantees) we'll never have a network with `id === -1`.

PR #222, opened 2017-11-01
WIP analysis via ngram modeling

boblannon:
This PR adds a few things that will make statistical token analysis possible and help users see potential differences in two different data sources.

### Frequency Counter
This is a very focused chunk of work. It should be an accumulator that looks at a series of documents one by one and returns the number of occurrences of each word (or multi-word sequence) found in the documents.

- input is a series of documents (strings)
- accumulate counts of one-word, two-word and three-word sequences

Helpful resources: TKTKTK

### Collocation Class
This is how the frequency counter should store its output in memory.

- Allows incrementing or querying the raw frequencies of any given n-gram.
- Can be serialized to some plain-text format
- Can be de-serialized from that same format
- Can apply an associativity measure to all of its members and return a ranked table
- Is composable with other instances of the same class

### Comparison Operation
There needs to be a function or class that can compare two collocation objects and highlight meaningful differences between them. Some possibilities:

- Are any high-frequency tokens found in A and not B?
    - what are some near-matches in B that are high-frequency?
    - this would catch, eg, stylistic differnces like "street" vs "st" vs "str"
- Do a pseudo-chi-square to identify which tokens (if any) are more associated with A or with B

### Associativity Measure(s)
Associativity Measures would work on a single collocation class and would describe how strongly associated words are to one another. This can be saved for a future iteration if we're concerned about time. We'll probably need to implement these ourselves, based on some nltk classes and pseudocode from a textbook Bob has.


boblannon:
context: https://github.com/mapbox/geocoding/issues/192

lib/analyze.js
```diff
@@ -0,0 +1,155 @@
...
+
+    if (argv.limit) {
+        argv.limit = parseInt(argv.limit);
+    }   else {
+        console.log('note: you can limit results/rows with --limit <number>');
+    }
+
+    var cc = argv.cc;
+    var type = argv.type;
+    var limit = argv.limit
```
ingalls: Mind replacing `var` with `let` or `const` depending on if you change these later on? I've banished var from the rest of the codebase and in doing so cleaned up several edge case bugs that persisted due to `vars` loose rules.

PR #230, opened 2017-11-03
Walkietalkie - Deubug Server Improvements II

KaiBot3000:
- [x] pull in web/index.html from master so we get nice alerts
- [x] generalizes alert color
- [x] use clipboard.js (from cdn)
- [x] copy map location
    - [x] add copy button to navbar
    - [x] add center of map as attribute when clicked
    - [x] copy to clipboard
- [x] copy feature json
    - [x] add copy button to code viewer
    - [x] when clicked, copy feature to clipboard
- [x] make lines transparent so we can see OSM road names

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@
 
 ## Version History
 
+### v12.6.4
+
+- :white_check_mark: Add broader language tests to tokenize
+
+### v12.6.3
+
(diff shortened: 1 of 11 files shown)


KaiBot3000:
- [x] limit precision so that we don't copy over the weird scientific notation version of coordinates

PR #219, opened 2017-10-31
Coverage for testcsv.js

lizziegooding:
### Context

Adds tests for `lib/testcsv.js`. Goal is to test how pt2itp is handling certain known edge cases using synthetic data.

### Summary of Changes
- [x] added test fixtures for Rhode Island addresses
- [x] fixed a bug in `lib/testcsv.js` that prevented csv files with less than 25 or less than a multiple of 25 lines from being geocoded
- [x] fixed typo in `lib/geocode.js`
- [x] added tests tracking `stdout` and errors written to output csv in `test/testcsv.js` using fastcsv and tape-spawn
- [x] removed .mbtiles files from the gitignore

### Next Steps
- [ ] preliminary review from @ingalls or @yhahn

PR #233, opened 2017-11-07
Add tests to test mode

lizziegooding:
## Context

Adds tests for test mode of pt2itp.

## Issues

I'm seeing the following undesired behavior with test mode:
- `meta.tokens`, the token option we pass to carmen when we geocode is undefined. It's set in `lib/map.js` but doesn't stick. This isn't an issue in map itself as it directly uses geocoder abbreviation tokens
- the test data shows that single digit address number addresses are being converted to single characters (i.e. `7 greenview rd` becomes `n greenview rd`) after being passed through `diacritics` and `tokenize`:
```
normalizedNtext = diacritics(tokenize(normalizedNtext, meta.tokens, true)[tokenKey].join(' '));
```

I created a unit test to try and isolate this behavior (`test/tokenize.diacritics.test.js`), but it behaves as expected
- The final stderr message is being printed before `diffName` has finished running, leaving a hanging progress bar that eventually times out

## Next steps

- [ ] merge in most recent changes from FreeRadicals
- [ ] start addressing/digging into issues above


lizziegooding:
## Summary of changes

- [ ] added test fixtures for testing test mode
- [ ] added unit test for handling `undefined`, `null`, and string replacer args
- [ ] added integration test for test mode that checks stderr and output csv
- [ ] change test mode stats so errors are shown as a percentage of total addresses, not total potential errors
- [ ] add checks within the map module that empty args aren't inserted into the PG meta table

## Resolved issues

**`meta.tokens` is undefined & single digit address numbers converted to characters**

Users should be able to omit the `tokens` argument when running the map module. However, this was creating an issue when testing test mode because values of `undefined` or `null` were being converted to strings by the `setMeta` method in `lib/index.js` when they were inserted into postgres. `lib/test.js` then retrieved this token and passed it as the `replacer` argument to `lib/tokenize.js`, which then replaced the address numbers n with the nth letter of `'undefined'` turning `6` into `i`, `7`, into `n` etc. I resolved this by throwing an error if the replacer wasn't an object in `lib/tokenize.js` and not inserting `null` or `undefined` values into postgres in the `lib/map.js` module.

 **final stderr message printed before test mode has finished**

This was because the `total` value of the error bar was being set to the number of addresses which didn't match to network rather than the number of addresses that didn't match to network and didn't match any of the aliases of the road network. Resolved by fixing the SQL query to include this second condition.

## Outstanding issues

The current test fails when comparing the expected and the actual `stderr` message. I've `diff`d the two messages and as far as I can tell they're exactly the same. I'm using the same syntax as I used successfully in https://github.com/ingalls/pt2itp/pull/219/files. Unfortunately the tape output from the error doesn't keep it's formatting, which I think may be tripping this up. I'm also not sure how the progress bar output may be affecting this-- it doesn't show up in the tape error message, so I don't think it should be included 🤷‍♀️ .

## Next steps

- [ ] review by @ingalls 
- [ ] fix `stderr` message issue-- @ingalls lmk if you have thoughts when you review

cc @ingalls @KaiBot3000 @sbma44 @yhahn

ingalls:
:taco: This looks great, instead of doing an exact match on the text can you just do a regex check? Like `String.match`. 

The reason these are likely failing is that to make progress bars/password prompts etc. Terminals have a delete character sequence. But the characters can't actually be deleted from the stream.

IE:
```
==>     |
<DELLINE>
===>   |
<DELLINE>
```

lizziegooding:
thanks for the tip, @ingalls remind :eyes: one more time and 🌮  if I'm good to merge into FreeRadicals?

PR #240, opened 2017-11-13
check for nonemptiness of 'other' val

sbma44:
Getting an error on `titlecase` when processing some WY data. I think that when both the primary text and the alternate text are empty/undefined, a runtime error was thrown. This change checks for that condition and returns an empty string in the rare cases where it occurs. Downstream processing should suppress the use of an empty-label feature.


sbma44:
no takers on review and confirmed this resolves my runtime error, yolo

Issue #241, opened 2017-11-15
proposal: improve testcsv results in a post-TIGER world


sbma44:
A conversation with @lizziegooding suggested two ways by which we can simplify our testing process:

- Lizzie discovered that some rows in our test CSVs don't have coordinates and wind up as `DIST` errors. Whoops! We should skip these.
- An emerging theme in TIGER-augmented work is an increase in `DIST` errors. We should check:  
  - the proportion of these that are interpolated results
  - if the synthetic TIGER points accurately reflect TIGER line geometry data
  - if the pt2itp/TIGER ➡️  carmen pipeline results in results that faithfully convey TIGER line geometry data
  - if our DIST problems simply reflect TIGER's imprecision and, if so, whether a looser distance threshold for DIST errors is appropriate for interpolated results

As a first pass @ingalls has suggested simply keeping the `DIST` and `DIST (ITP)` metrics separate from one another. I'll start on a PR to do this and handle the bad CSV row issue shortly.

cc @aarthykc @aaaandrea @KaiBot3000

sbma44:
we haven't had reason to question TIGER points' validity (though we might regenerate them at a lower density). For now we're not adjusting the DIST threshold. However we _have_ landed some PRs that expand testcsv verbosity and give us more insight into what's going on. closing for now.

PR #244, opened 2017-11-15
ITP testcsv granularity

sbma44:
implements ideas from https://github.com/ingalls/pt2itp/issues/241

- skip testcsv rows that are missing valid geometry
- separate ITP distance fails from point distance fails
- collect data on distance fails to give a better sense of ranges (to assist evaluation of whether to adjust the ITP dist threshold)

### DIFF
--- a/lib/testcsv.js
+++ b/lib/testcsv.js
@@ -88,7 +88,7 @@ function test(argv, cb) {
     /**
      * Test for numeric-ness
      * @param {Object} n potential number
-     * @return {Boolean} isnumber
+     * @return {boolean} isnumber
      */
     function isNumber(n) {
         return !isNaN(parseFloat(n)) && isFinite(n);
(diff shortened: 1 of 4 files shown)


lib/testcsv.js
```diff
@@ -84,6 +85,15 @@ function test(argv, cb) {
 
     let buffer = [];
 
+    /**
+     * Test for numeric-ness
+     * @param {Object} n potential number
+     * @return {Boolean} isnumber
```
ingalls:
```
type Boolean found, boolean is standard
```

sbma44:
Expanding on the above, this adds output like the following:

```
DIST (ITP)                     157 ( 63.1% of errors | 10.4% of total addresses)
DIST (STREET)                   64 ( 25.7% of errors |  4.2% of total addresses)
NO RESULTS                      23 (  9.2% of errors |  1.5% of total addresses)
TEXT                             5 (  2.0% of errors |  0.3% of total addresses)

ok - 249/1508 (16.5%) failed to geocode

DIST statistical breakdown
DIST (ITP) / mean: 55.20 / median: 3.52 / skew: 1.95 / standard dev: 89.53
```

PR #245, opened 2017-11-15
track how many results are ITP

sbma44:
Satisfy your morbid curiosity: how many points are coming back versus interpolated results? _Should_ only be useful in testcsv mode, but could serve as an indicator of problems in vanilla test mode.


sbma44:
cc @boblannon @aaaandrea @aarthykc @KaiBot3000 @ingalls @lizziegooding

PR #248, opened 2017-11-16
End the1 percent

ingalls:
Streets that had duplicate addresses were not getting passed into `explode#split` which meant they could be of any length. This standardizes the output format of `cluster#break` to simply be an array in the same format as `segs` in `split`. this can then be easily passed to `explode#split` and then `split#distribute`


ingalls:
This section of code is both untested and currently broken: https://coveralls.io/builds/14241049/source?filename=lib%2Fcluster.js#L130

PR #254, opened 2017-11-29
analyze.js - Performing SQL copy commands in bulk

boblannon:
Inserting one record at a time turned out to be slow and was also pretty disastrous. It eventually raises `RangeError: Maximum call stack size exceeded` from `d3-queue`, which seems to have a hard-coded limit on the number of tasks that can be deferred.

And anyway it was ugly.

This refactors things so that we populate bigram and unigram tables with a single `COPY FROM` and produce outputs using `COPY TO`.

One thing is still unresolved, though, and it's driving me insane: for some reason, the last `copyOutput` task deferred in `compareAddressAndNetwork` tends not to produce any actual output for real-world-sized data. It doesn't cause tests to fail, but the fixture data is super tiny.  That suggests a race condition, but I can't figure out where it would be.


lib/analyze.js
```diff
@@ -559,55 +566,38 @@ function compareAddressAndNetwork(argv, pool, callback) {
...
-    for (let i=0; i< typeOrderPairs.length; i++) {
-        let [type, order] = typeOrderPairs[i];
-        outputPath = `${outputPrefix}-${type}-${order}.csv`;
-        popQ.defer(copyOutput, type, order, outputPath);
     }
 
+    popQ.defer(copyOutput, 'address', 'bigram');
+    popQ.defer(copyOutput, 'address', 'unigram');
+    popQ.defer(copyOutput, 'network', 'bigram');
+    popQ.defer(copyOutput, 'network', 'unigram');
```
boblannon: when using realworld-sized data (like a whole US state), whichever task gets deferred last is the one that doesn't finish.  as-is, the network-unigram CSV file contains only a header.  if you swap the lines so that network-bigram is last, then that CSV file is the one with only a header, and the network-unigram CSV is fine.

boblannon: ah, okay, this looks like it's related to [a quirk in `pg-copy-streams`](https://github.com/brianc/node-pg-copy-streams/issues/61), where the `'end'` event in the copy stream doesn't guarantee that the stream being piped to has flushed.

PR #255, opened 2017-11-29
remove pretensions to multilingual capitalization support

sbma44:
Automatic titlecasing behavior kicks in when we don't detect adequate capitalization signals from either the network or address data. This should occur rarely. When it does, we have been using English title-case behavior: capitalizing every word except a blacklist of prepositions and articles. Although English rules were always used, code paths had been stubbed out to allow non-English support in the future.

Some research shows that the dominant non-English pattern for capitalization is "first letter plus proper nouns" (with some variations around adjectival forms of proper nouns). Detection of proper nouns is difficult, as is making sure the capitalized address data meshes with other data in the geocoding response (which probably includes city names, etc).

Tackling support for non-English capitalization rules is therefore a big quest and one that's out of scope for right now. This PR removes some of the interface elements that might lead users to incorrectly think that non-English capitalization rules are supported. The hardcoding to English is made more explicit in `titlecase.js`.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -22,7 +22,7 @@ class Cluster {
         if (opts.label)
             this.label = require('./label/' + opts.label)();
         else
-            this.label = require('./label/titlecase')({ language: 'en' });
+            this.label = require('./label/titlecase')();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


sbma44:
cool -- there is a titlecase test but it needed to be updated to reflect these changes

PR #266, opened 2017-12-07
add fr tokens; handle streets w/ numeric names but no suffixes

sbma44:
Two changes, the first set being a minor one:
- add `Boul` token for `Boulevard`; `Aut` token for `Autoroute`; both are scoped to French

The second is more substantial. In cases where both the input network and address data look like `61 Avenue`, queries for `61st Avenue` might fail (depending on your input tokenization scheme). There are other ways to handle this but @ingalls and I decided that adding a lower-priority synonym is the one we would pursue for now, turning this hypothetical feature's text into `61 Avenue,61st Avenue`. 

This PR accomplishes this by 
- creating additional clones of network segments via `lib/map/osmium.js` for appropriate features in the US/CA/GB country contexts
- adding the notion of `priority` to network segments and clusters, which is in turn used to prioritize the order of synonyms during the `collapse()` operations in order to ensure that the source data's label remains canonical


sbma44:
tagging @ingalls for review because it might make him feel guilty about the osmium.js refactor he said he's about to land

aarthykc:
👌

PR #278, opened 2017-12-14
Name Alts

ingalls:
## Intent

This PR builds out better support for standardized synonyms on entry to pt2itp.

Networks and Address points can both have synonyms, whether it be a different language or simply another name/way of representing the name. At the moment we support synonyms on network features by simply duplicating the network feature and then merging similar streets back together in the collapse step. We can't support address synonyms in this way as one of the address would get matched and the other potentially orphaned when they they should be matched together.

This changes the way text is stored internally to be an array of synonyms for addresses, and an identical text object for networks (retaining the merge code)

```
{
    display: "Display name, formerly _text",
    tokenized: "Tokenized name, formerly text",
    tokenless: "Tokenless name, formerly text_tokenless"
    priority: Int (0 default)
}
```

This solves several problems, opening up better behavior in linking code as well as giving a significant amount of flexibility to the final text field in displaying the best field. Right now this field is not entirely deterministic, with some features being named `NC 33, Highway 33` and others `Highway 44, NC 44`.

## Cons

QGIS doesn't support display of the `JSONB` type (which is required to have indexes). To view the contents in qgis we will need to run a script after itp creation which creates temporary debug columns of type `TEXT` for QGIS.

## Relevant Issues:

Ref: https://github.com/ingalls/pt2itp/issues/276


sbma44:
How about [setting up some views](https://www.postgresql.org/docs/current/static/sql-createview.html) for QGIS rather than manipulating the tables manually?

ingalls:
@sbma44 I hadn't thought of that, that's a wonderful idea!

PR #281, opened 2017-12-18
Diffname speed

boblannon:
It seems like https://github.com/ingalls/pt2itp/pull/272 has failed to speed things up sufficiently.  here's an example from the `br_sp` build, which has been running for almost 4 days:

```sql
EXPLAIN SELECT COUNT(*) FROM
  (SELECT * FROM address a WHERE netid IS NOT NULL) a
  WHERE a._text NOT IN
  (SELECT DISTINCT(UNNEST(REGEXP_SPLIT_TO_ARRAY(n._text, ','))) AS uniqtext FROM network_cluster n);
```
```
                                                QUERY PLAN
-----------------------------------------------------------------------------------------------------------
 Aggregate  (cost=1800887388976.58..1800887388976.59 rows=1 width=8)
   ->  Seq Scan on address a  (cost=45879.52..1800887375923.34 rows=5221295 width=0)
         Filter: ((netid IS NOT NULL) AND (NOT (SubPlan 1)))
         SubPlan 1
           ->  Materialize  (cost=45879.52..312425.84 rows=15802200 width=32)
                 ->  Unique  (cost=45879.52..125391.84 rows=15802200 width=32)
                       ->  Sort  (cost=45879.52..46525.24 rows=258287 width=32)
                             Sort Key: (unnest(regexp_split_to_array(n._text, ','::text)))
                             ->  Seq Scan on network_cluster n  (cost=0.00..16480.31 rows=258287 width=32)
(9 rows)
```

This PR removes the query for `totalDiffNames` altogether.  As it turns out, it was only used for the progress bar.  In the future, if we'd like to add the progress bar back, we can try to dream up a better query.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -117,6 +117,7 @@ function test(argv, cb) {
     }
 
     if (argv.limit) argv.limit = parseInt(argv.limit);
+    const limitClause = (argv.limit ? ' LIMIT ' + argv.limit : '');
 
     const pool = new pg.Pool({
         max: 10,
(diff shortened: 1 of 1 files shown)


boblannon:
yeah, it's a little strange to me, too. that's the pattern of the other tests as well:

```javascript
client.query(`<QUERY TO GET NUMBER OF RELEVANT RECORDS>`, (err, res) => {
        if (err) return cb(err);

        const cursor = client.query(new Cursor(`<QUERY FOR TEST USE>`));

        const bar = new Prog('ok - Testing [:bar] :percent :etas', {
            complete: '=',
           incomplete: ' ',
           width: 20,
           total: argv.limit || res.rows[0].count++
       });
...
```

PR #270, opened 2017-12-11
Change context of csvStream.resume

lizziegooding:
Testcsv mode has been failing with the following error:
```
_stream_readable.js:718
  var state = this._readableState;
                  ^

TypeError: Cannot read property '_readableState' of undefined
    at Readable.resume (_stream_readable.js:718:19)
    at Queue.addrQ.awaitAll [as _call] (/Users/lizziegooding/pt2itp/lib/testcsv.js:170:26)
    at maybeNotify (/Users/lizziegooding/pt2itp/node_modules/d3-queue/build/d3-queue.js:120:7)
    at /Users/lizziegooding/pt2itp/node_modules/d3-queue/build/d3-queue.js:91:12
    at Function.isPass (/Users/lizziegooding/pt2itp/lib/geocode.js:70:12)
    at c.geocode (/Users/lizziegooding/pt2itp/lib/testcsv.js:150:33)
    at finalize (/Users/lizziegooding/pt2itp/node_modules/@mapbox/carmen/lib/geocode.js:396:16)
    at Queue._call (/Users/lizziegooding/pt2itp/node_modules/@mapbox/carmen/lib/verifymatch.js:209:9)
    at maybeNotify (/Users/lizziegooding/pt2itp/node_modules/d3-queue/build/d3-queue.js:120:7)
    at Queue.awaitAll (/Users/lizziegooding/pt2itp/node_modules/d3-queue/build/d3-queue.js:51:5)
```
This was introduced in https://github.com/ingalls/pt2itp/pull/264/files when the geocodeBuffer function was created to move functionality outside of the csv parsing function. The issue is that when `csvStream.resume` is called within the d3.queue it has an undefined context (`this`) and therefore errs out. Passing`csvStream.resume` as an argument in an anonymous function fixes this.

PR #282, opened 2017-12-18
Add tests for testcsv inputs >= 25 entries

lizziegooding:
As a follow up to https://github.com/ingalls/pt2itp/pull/270 which fixed a bug introduced in https://github.com/ingalls/pt2itp/pull/264, update test fixtures for testcsv tests to check that csv inputs with >= 25 rows are successfully geocoded. The current unit tests never reach this conditional statement: https://github.com/ingalls/pt2itp/pull/270/files#diff-7aa9bba0f48549762eaa30f3aada66cfL113. To be totally sure, I ran the tests on this branch against `v13.6.4` of pt2itp and verified that these tests would have failed, thus 🎏 ing the issue.

cc @yhahn


yhahn:
👍

PR #284, opened 2017-12-20
Update unit tests for map/strip-unit.js to accept street arrays

lizziegooding:
As part of https://github.com/ingalls/pt2itp/pull/278, updates map/strip-unit.js and test/map.strip-unit.test.js to accept GeoJSON with a street property that can be either a string or a properly formatted array, e.g.:

```
properties: {
            number: 1,
            street: [{
                display: 'Main St',
                priority: 0
            }]
        }
```

## Next steps
- [x] Update README.md
- [ ] review by @ingalls 
- [ ] merge

### DIFF
--- a/lib/map/strip-unit.js
+++ b/lib/map/strip-unit.js
@@ -10,22 +10,29 @@ function map(feat, context) {
     if (feat.geometry.type !== 'Point') return new Error('Feat must be a Point geom');
     if (!feat.properties) return new Error('Feat must have properties object');
     if (!feat.properties.number) return new Error('Feat must have number property');
-    if (!feat.properties.street) return new Error('Feat must have street property');
-    if (typeof feat.properties.street === 'string' && !feat.properties.street.trim().length) return new Error('Feat must have non-empty street property');
 
-    if (Array.isArray(feat.properties.street)) {
-        if (feat.properties.street.length <= 0) return new Error('Feat must have non-empty street property');
-        if (!feat.properties.street.some(syn => Object.keys(syn).includes('display')) || !feat.properties.street.some(syn => Object.keys(syn).includes('priority')) || !feat.properties.street.some(syn => Object.keys(syn).length === 2)) return new Error('Every synonym object in street array must contain only display and priority properties');
+    if (!feat.properties.street) return new Error('Feat must have street property');
+    if (typeof feat.properties.street === 'string') {
+        if (!feat.properties.street.trim().length) return new Error('Feat must have non-empty street property');
+
+        feat.properties.street = [{
+            display: feat.properties.street,
+            priority: 0
+        }];
+
+    } else if (Array.isArray(feat.properties.street)) {
+        if (!feat.properties.street.length) return new Error('Feat must have non-empty street property');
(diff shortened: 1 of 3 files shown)


lib/map/strip-unit.js
```diff
@@ -11,9 +11,18 @@ function map(feat, context) {
     if (!feat.properties) return new Error('Feat must have properties object');
     if (!feat.properties.number) return new Error('Feat must have number property');
     if (!feat.properties.street) return new Error('Feat must have street property');
-    if (!feat.properties.street.trim().length) return new Error('Feat must have non-empty street property');
+    if (typeof feat.properties.street === 'string' && !feat.properties.street.trim().length) return new Error('Feat must have non-empty street property');
 
-    if (!feat.geometry.coordinates || !feat.geometry.coordinates instanceof Array || feat.geometry.coordinates.length !== 2) return new Error('Feat must have 2 element coordinates array');
+    if (Array.isArray(feat.properties.street)) {
+        if (feat.properties.street.length <= 0) return new Error('Feat must have non-empty street property');
```
ingalls: An array length can never be `<` 0 - Won't hurt anything though!

ingalls: An array length can never be `<` 0 - Won't hurt anything though!

lib/map/strip-unit.js
```diff
@@ -11,9 +11,18 @@ function map(feat, context) {
...
+        if (!feat.properties.street.some(syn => Object.keys(syn).includes('display')) || !feat.properties.street.some(syn => Object.keys(syn).includes('priority')) || !feat.properties.street.some(syn => Object.keys(syn).length === 2)) return new Error('Every synonym object in street array must contain only display and priority properties');
+
+        for (let i = 0; i < feat.properties.street.length; i++) {
+            if (typeof feat.properties.street[i].display !== 'string' || isNaN(feat.properties.street[i].priority)) return new Error('Display property must be a string and priority property must be a number');
```
ingalls: `isNaN` doesn't do quite what you think in this context. `NaN` is actually a javascript type in this case (like an integer, string, array, etc) 

Instead you will want to cast the variable to a numeric type and test if it is `NaN`.

IE: `isNaN(Number('avc'))` will return true

lib/map/strip-unit.js
```diff
@@ -35,12 +44,14 @@ function map(feat, context) {
 
     if (feat.properties.number.length > 10) return new Error('Number should not exceed 10 chars');
 
-    feat.properties.street = [{
-        display: feat.properties.street,
-        priority: 0
-    }];
-
+    if (typeof feat.properties.street === 'string') {
```
ingalls: Can you add this to the if Array flow?

ie:
```
if Array {
   - Do array stuff
} else if (string)
  do string stuff
}
```

This will avoid every feature running through 2 if statements.

Current Flow:
```
if Array {
do stuff
}

if string {
 do stuff
}
```

test/map.strip-unit.test.js
```diff
@@ -64,7 +106,7 @@ test('Strip-Unit', (t) => {
         },
         geometry: {
             type: 'Point',
-            coordinates: [1000, 1000]
+            coordinates: [1000, 0]
```
ingalls: This looks strange, a null value should fail, not be cast to 0, welcome to null island! 🎉

lib/map/strip-unit.js
lizziegooding: Sounds good! I was following the pattern from https://github.com/ingalls/pt2itp/pull/284/files#diff-3c842f03a7487f8a3ff4009eef957ff4L18, want me to change that as well?

PR #285, opened 2017-12-22
Moar unit tests for map/strip-unit.js to err on non string/array streets

lizziegooding:
## Context

When running the full US index with the new `addralts` branch, @ingalls found the following error in the lib/label/titlecase.js module:
```
/usr/local/src/mapbox-places-address/node_modules/pt2itp/lib/label/titlecase.js:22
15:12:43
if (!name.display.trim()) return false;
15:12:43
^
15:12:43
TypeError: Cannot read property 'trim' of undefined
15:12:43
at _.uniqBy.filter (/usr/local/src/mapbox-places-address/node_modules/pt2itp/lib/label/titlecase.js:22:30)
15:12:43
at Array.filter (native)
```
It looks like this was happening because `name.display` was `undefined`. In https://github.com/ingalls/pt2itp/pull/284, I updated lib/map/strip-unit.js to accept either string or array `street` properties, but didn't check for or throw an error when `street` was any other type. This resulted in the output `street` property not being modified and therefore not containing an array of objects with `display` or `priority` properties.

## Summary of changes
- [ ] added error handling for non-string or array street properties
- [ ] added unit test for non-string or array street properties in 
- [ ] removed duplicate titleCase function in lib/label/titlecase.js

## Next steps
- [ ] review
- [ ] merge into `addralts`
- [ ] update pt2itp version in mapbox-places-address branch
- [ ] re run us index on watchbot

cc @ingalls


ingalls:
🌮

PR #286, opened 2017-12-28
Sanitize newline and carriage returns when writing psv's

yhahn:
Ran into some network street names that had a few more psv-unsafe characters that needed sanitizing:

- `\n` - replaces with a space
- `\r` - replaces with a space

This branch also adds a test for garbage input to confirm it gets cleaned up for psv output.


ingalls:
Very good catch - 🌮

yhahn:
Thanks for the review @ingalls !

PR #287, opened 2018-01-02
Bad Synonyms

ingalls:
Add list of tokens that are less desirable to be picked as the main display label.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -194,7 +194,7 @@ class Cluster {
 
             INSERT INTO address_cluster (name, geom, netid)
                 SELECT
-                    JSON_Agg(a.name||('{ "freq": '::TEXT||ST_NPoints(geom)||'}')::JSONB ORDER BY ST_NPoints(geom)),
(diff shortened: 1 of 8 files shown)


ingalls:
New Passrates:

```
DIST                           299 ( 33.4% of errors |  3.1% of total addresses)
DIST (ITP)                     303 ( 33.8% of errors |  3.1% of total addresses)
DIST (STREET)                   47 (  5.2% of errors |  0.5% of total addresses)
NO RESULTS                      31 (  3.5% of errors |  0.3% of total addresses)
TEXT                           216 ( 24.1% of errors |  2.2% of total addresses)

ok - 896/9779 (9.2%) failed to geocode
ok - 303/1128 (26.9%) ITP results failed to geocode
```

Vs: Old Passrates: `12.3%` (Don't have a full record)

ingalls:
This PR Introduces two methods in which to bias results to the correct network name. 

The first is a list of non-preferred street tokens: IE: 'ext', 'unit', 'exit' etc which suggest an incorrect name.

The second is a `freq` value that is used to determine the best match based on the frequency of a given name. Previously `Grand Valley Dr` (1 addr) would always beat `Grand Valley` (307 addr) as it was longer and assumed to be unabbreviated.

PR #290, opened 2018-01-06
tokenize createGlobalReplacer in test

aaaandrea:
Goal: use tokenize to tokenize strasse as end results expect. 

#### Next steps
- [x] recajigger the thingamawhatsit to the whatchamacallit
- [x] get tests to pass 

     currently running into this: 
     ```
     # Return correct std.err message
     not ok 3 "ok - beginning match ok - beginning unmatch ok - beginning diff name ERROR TYPE                     COUNT ----------------------------------------------------------------------------------- NOT MATCHED TO NETWORK 1 ( 100.0% of errors | 9.1% of total addresses) ok - 0/11 (0.0%) failed to           geocode ok - 0/0 (NaN%) ITP results failed to geocode DIST statistical breakdown ----------------------------------------------------------------------------------- " did not match /NAME      MISMATCH \(SOFT\)\s+1 \( 50\.0% of errors \| 9\.1% of total addresses\)/
       ---
         operator: ok
              expected: true
              actual:   false
              at: Socket.onEnd (/usr/local/src/pt2itp/node_modules/tape-spawn/index.js:106:14)
     ```


test/test-de.test.js
```diff
@@ -0,0 +1,119 @@
...
+        t.end();
+    });
+    console.log('*** de Drop/init database ended');
+});
+
+// loads address and network data into postgres
+test('load address and network de files', (t) => {
+    const popQ = new Queue(1);
+    popQ.defer((done) => {
+        worker({
```
yhahn: @aaaandrea it looks like these files aren't committed to the repo:

```
test/fixtures/test-de/address.geojson
test/fixtures/test-de/network.geojson
```

I think it might be because there is an entry in .gitignore for *.geojson (https://github.com/ingalls/pt2itp/blob/master/.gitignore#L6) so that they won't get added automatically. You may have to explicitly add them from your local repo with

```
git add -f test/fixtures/test-de/address.geojson
git add -f test/fixtures/test-de/address.geojson
```

I think this might be what's hanging the circle testrun? (Passes also hang for me locally as I don't have these files...)

aaaandrea:
Breakdown of changes and how this works. 
- [x] test.js changes implement global tokens. 
- [x] geocode.js changes allow test analysis to also take this into account. 
- [x] tests no longer rely on static results, but iterate based on what's in master. 

side thing I'm slipping in:
- [x] I removed a log about limiting the analyze script. It's no longer necessary since the script is super fast. I moved it to a comment, in the off chance somebody still wants to do that.

test/testcsv.test.js
```diff
@@ -6,7 +6,7 @@ const path = require('path');
 const carmenIndex = path.resolve(__dirname, './fixtures/test-ri/index/us_ri-address-both-0d603c2a171017011038-0d603c2a39.mbtiles');
 const input = path.resolve(__dirname, './fixtures/test-ri/address.csv');
 const output = '/tmp/testcsv-ri.err';
-const config = path.resolve(__dirname, './fixtures/test-ri/carmen-config.json');
+const config = path.resolve(__dirname, './fixtures/test-ri/carmen-config2.json');
```
lizziegooding: I'm curious, can you say more about what these two different carmen-configs do and why we use `carmen-config.json` for pt2itp test mode tests and `carmen-config2.json` for testcsv mode?

test/geocode.test.js
```diff
@@ -22,7 +22,8 @@ test('geocode#isPass', (t) => {
             tokens: tokens.createReplacer(['en'])
         },
         function(err, res) {
-            t.error(err)
+            t.error(err);
+            console.log(res);
```
lizziegooding: mind removing this :) ?

test/testcsv.test.js
aaaandrea: I opened a ticket which addresses the need for some changes [here](https://github.com/ingalls/pt2itp/issues/303). 

Decided that testcsv tests currently cover what they are supposed to cover using the old config (carmen-config2) which is a complete json. 

When we have more time though we should definitely go back and clean this up.

Issue #291, opened 2018-01-08
strip-unit returns mismatched error message on blank `number`


trescube:
When the input to strip-unit has a `number` property with length greater than 0 and is entirely whitespace, an error about units is returned:

```javascript
> stripunit({
...     geometry: {
.....       type: 'Point',
.....       coordinates: [
.....         12.121212,
.....         21.212121
.....       ]
.....     },
...     properties: {
.....       number: ' \t ',
.....       street: 'main street'
.....     }
...   });
Error: Feat is not a supported address/unit type
    at map (/Users/sprinter5/Mapbox/mapbox-places-address/node_modules/pt2itp/lib/map/strip-unit.js:50:178)
```

The `street` property is trim'd and checked for zero-length, the same should probably happen for `number`.

lizziegooding:
Fixed in https://github.com/ingalls/pt2itp/pull/293

PR #293, opened 2018-01-09
Handle empty string and non-string or num address numbers

lizziegooding:
## Context

Addresses https://github.com/ingalls/pt2itp/issues/291 where given a GeoJSON feature with a non-empty whitespace `number` property, we returned `Error: Feat is not a supported address/unit type` rather than a `number` property error.

While I was looking into this, I added error handling for non string or number `number` properties and moved all address number parsing code together so if follows a similar format to how we're parsing the `street` property.

## Summary of changes
- [x] added a check for whitespace `number` property values and a unit test
- [x] added a check for non string or number `number` property values and a unit test
- [x] moved all code that handles the `number` property together

## Next steps 
- [ ] review by @trescube and @ingalls 
- [ ] merge & release

### DIFF
--- a/lib/map/strip-unit.js
+++ b/lib/map/strip-unit.js
@@ -16,7 +16,7 @@ function map(feat, context) {
     } else if (typeof feat.properties.number === 'number') {
         feat.properties.number = String(feat.properties.number);
     } else {
-        return new Error('Feat must have a string or number number property');
+        return new Error('Feat must have a string or numeric number property');
     }
 
     feat.properties.number = feat.properties.number.toLowerCase();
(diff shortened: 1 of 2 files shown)


lib/map/strip-unit.js
```diff
@@ -9,7 +9,28 @@ function map(feat, context) {
...
     if (feat.geometry.type !== 'Point') return new Error('Feat must be a Point geom');
     if (!feat.properties) return new Error('Feat must have properties object');
+
     if (!feat.properties.number) return new Error('Feat must have number property');
+    if (typeof feat.properties.number === 'string') {
+        if (!feat.properties.number.trim().length) return new Error('Feat must have non-empty number property');
+    } else if (typeof feat.properties.number === 'number') {
+        feat.properties.number = String(feat.properties.number);
+    } else {
+        return new Error('Feat must have a string or number number property');
```
trescube: The word numeric would make this look less like a typo.

lizziegooding: 👍 good call!

PR #300, opened 2018-01-19
Use max JS integer size when setting ID

mattficke:
PT2ITP currently only uses 31 bits for IDs, which leads to ID collisions with some degree of frequency. This updates the ID script to use [`Number.MAX_SAFE_INTEGER`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), an alias for 2^53 - 1 that's a little easier for humans to understand and read.

### DIFF
--- a/lib/post/id.js
+++ b/lib/post/id.js
@@ -6,6 +6,8 @@
 function post(feat) {
     if (!feat) return feat;
 
+    // Use all 53 bits we have available when setting ID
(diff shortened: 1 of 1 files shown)


ingalls:
🌮

Once merged please follow the steps in CONTRIBUTING.md to update the changelog, version, & public.

PR #309, opened 2018-01-25
analyze.js tweaks: use tokenized field

boblannon:
The results of the `analyze` command are meant to be useful for identifying tokens that might prevent matches between addresses and networks.  Currently, though, it is doing a statistical analysis of the `display` properties in `address.name` and `network.name`. It ends up, however, reporting a lot of tokens that are already handled by our tokenization process, which collapses variants of things like address suffixes (eg "Road" vs "Rd").

A more informative analysis would describe the `tokenized` property of these name fields, which benefits from token normalization and is the field actually used when matching address and network records.

### DIFF
--- a/lib/analyze.js
+++ b/lib/analyze.js
@@ -441,7 +441,7 @@ function compareAddressAndNetwork(argv, pool, callback) {
                 FROM
                     bigram_comparison
                 WHERE
-                    zscore > 1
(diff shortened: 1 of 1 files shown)


ingalls:
@boblannon am I ok to close? Or should this be wrapped up/merged?

PR #313, opened 2018-02-02
Network Synonyms

ingalls:
## Context

As of v14.x, PT2ITP uses JSON array to represent all text on input addresses and network features.

This JSON format is persisted through to the final merged output for addresses and makes debugging how synonyms are handled internally rather straightforward.

Unfortunately this is not the case for Network Segments. Unlike an address which is a discrete geometric primitive, network segments must be combined together to form the full street. Synonyms make this difficult as a complex and loose set of rules must be used to determine when one street starts and another ends.

```
-------------------------------------
Main St / CA2     | CA 2
```

To simplify this merge operation, the network segments are only allowed to have a single name. Input network features that have multiple names are exploded into duplicate geometries, each with the distinct name. These geometries are then clustered on name & geographic proximity and where they are nearly equal merged together.

In the above example this would result with a final output of:

```
-------------------------------------
CA 2

----------------------
Main St
```

This PR explores what it would mean to use the same Array of text synonyms that is 
used on addresses for network segs.


apendleton:
Not sure I understand that last bit. We do consider synonyms in `matching_text` / `matching_place_name`: https://github.com/mapbox/carmen/blob/master/lib/util/ops.js#L235-L251

ingalls:
Per convo and subsequent tests - fix for waffle indexed features: https://github.com/mapbox/carmen/pull/685

PR #353, opened 2018-06-28
Better Cluster Breaking

ingalls:
After addresses are matched to their corresponding street segment, they are joined together to form the longest continuous segments possible. 

ie:

```
   seg 1    seg 2     seg 3
---------|---------|---------
```
would be joined to form
```
seg 1
-----------------------------
```

This allows a more consistent output (and better ITP segments), irrespective of segment formatting from the upstream street network provider.

Often highways and main urban roads will have an address reset somewhere on the road.

ie
```
1 2 3 4 5 1 2 3 4 5
------------------------
````

these often occur at the borders of city/municipal boundaries or zip codes. We have long handled these with the `cluster#break algorithm, which detects duplicate addresses that do not refer to the same physical street address. Once detected the feature is broken into 2 distinct features.

The current algorithm can only handle a single break, meaning streets with multiple address resets are not handled.

ie:

```
1 2 3 4 5 1 2 3 4 5
------------------------
```
would be handled correctly, while the below example would not be handled

```
1 2 3 4 5 1 2 3 4 5 1 2 3 4 5
-----------------------------------
```

This PR resolves this problem by 
- rewriting the cluster#break algorithm from the ground up to assume that multiple break points can occur in any given cluster.
- Add exhaustive tests for single/multiple address cliffs, single/multiple address humps, and combinations of the two

### CODE REVIEW

**Comment by ingalls** [comment]
Examine Encoded Address Values:

```
  { num: '\u0000\u0000undefined', output: true },                                                     
```
Several hundred addresses  appear to be malformed in some way.

ingalls:
Examples of house number reset:

screenshot from 2018-07-03 12-43-23

ingalls:
screenshot from 2018-07-11 08-30-41

Still seeing some errors with live data

PR #396, opened 2019-02-06
Increment the number of nearest networks to match per address

samely:
Some points are not being matched properly due to the limit of 10 nearest networks that are possible matches per address point.


samely:
The issue was fixed with the last refactoring of pt2itp.
Closing.

PR #400, opened 2019-02-12
Evaluate the nearest network to assign a high score

samely:
I've been playing around with the way we assign score to the network:

Found cases of mismatching due to the way we score networks.

**i.e.**
`e some ave` matching `e another ave` (far)  instead of matching `some avenue` (nearest)

I considered evaluating the nearest network before the other ones and which have just 2 characters different and gave a high score. 

@ingalls


samely:
Need a token refactoring.

PR #407, opened 2019-02-18
Better escape

ingalls:
The `CSV` Format of PG Copy has far better behavior for sensible escaping in JSON data.


boblannon:
does this resolve the escaping problems that i was trying to fix with https://github.com/ingalls/pt2itp/pull/405?

ingalls:
@boblannon Yes, it should

PR #411, opened 2019-02-22
Speed Quest

ingalls:


ingalls:
<h2 align=center>Paused</h2>

I'm pausing this work for the time being as to achieve the speed goals I'm going to have to do a pivot on how this works under the hood

- Parallel should take a GeoStream, context, & error log
- Threads should be fed GeoJson objects from the GeoStream
- Leaving processing & expensive work performed in AddrStream in the worker thread

PR #412, opened 2019-02-25
Performance Quest

boblannon:
## Progress To Date
- Added timed runs to each step in `map` mode
- Rewrote orphan query to run significantly faster

## PR Changes
- Rewrite Match query to run ~50% faster
- Rewrite Split query to run faster

### DIFF
--- a/lib/map.js
+++ b/lib/map.js
@@ -295,7 +295,7 @@ function main(argv, cb) {
      * Find closest Segment in network for each address point in matched cluster and generate ITP output
      */
     function splitter() {
-        split.prepare(pool, (err, res) => {
+        split.Split.prepare(pool, (err, res) => {
             pool.query(`SELECT id FROM network_cluster WHERE address IS NOT NULL;`, (err, res) => {
                 if (err) return cb(err);
 
--- a/lib/map.js
+++ b/lib/map.js
@@ -378,7 +378,7 @@ function main(argv, cb) {
                     child.send(opts);
                 }
             });
-        
+
         });
     }
 
--- a/lib/map.js
+++ b/lib/map.js
@@ -295,87 +295,90 @@ function main(argv, cb) {
      * Find closest Segment in network for each address point in matched cluster and generate ITP output
      */
     function splitter() {
-        pool.query(`SELECT id FROM network_cluster WHERE address IS NOT NULL;`, (err, res) => {
-            if (err) return cb(err);
+        split.prepare(pool, (err, res) => {
+            pool.query(`SELECT id FROM network_cluster WHERE address IS NOT NULL;`, (err, res) => {
+                if (err) return cb(err);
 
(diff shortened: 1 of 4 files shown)


ingalls:
@boblannon worked on a few things that I've pushed up here
- Updated w/ master which now includes time counts for each step
- Moved create table to it's own function as before it was called once per network feature

ingalls:
@boblannon seeing problems with the query - exhibited via the `test/split.splitfeat.test.js` test

```
    expected: |-                                                                                         
      { 1: { id: 1, number: '100', props: { postcode: '1' }, output: false }, 2: { id: 2, number: '200', 
props: { postcode: '2' }, output: false }, 3: { id: 3, number: '300', props: { postcode: '3' }, output: $
alse }, 4: { id: 4, number: '400', props: { postcode: '4' }, output: true }, 5: { id: 5, number: '500', $
rops: { postcode: '5' }, output: true }, 6: { id: 6, number: '600', props: { postcode: '6' }, output: tr$
e } }                                                                                                    
    actual: |-                                                                                           
      { null: { id: null, number: 'null', props: {}, output: null } }   
```

boblannon:
sigh! what a journey. as it turned out one of the major things plaguing the `split` query was the lack of indexes on fields that were being (implicitly) used in joins.  The changes to the query are now:
- prepare step:
  - creating a join table (`address_cluster_id_to_z`) to bridge `address_cluster` and `address`
  - add indexes for `interesections.a_id`, `intersections.b_id`, and `network_cluster.address`
- select query:
  - move the `address_props` out of a subselect into an actual join
  - group by the derived `name`, `network_cluster.geom` and `address_cluster.geom`

I did some `EXPLAIN ANALYZE` profiling, using `us_dc`.  I obtained the id of a network cluster that happened to have a lot of addresses and intersections using this query:

```sql
select name, array_agg(distinct nc_id), count(*) from (
    select nc.id as nc_id, nc.name || ac.name as name, nc.geom as network_geom, ac.id, ac.geom as address_geom, a.id, i.id
    from network_cluster nc join address_cluster ac on (nc.address = ac.id)
                            join address_cluster_id_to_z aciz on (ac.id = aciz.id)
                            left join address a on a.id = aciz.z
                            left join intersections i on (i.a_id = nc.id or i.b_id = nc.id)
    where nc.id < 100) x group by name, network_geom, address_geom
order by count(*) desc;
```

the top network had 176 addresses. 

here's the top line of `EXPLAIN ANALYZE` for the old Query with indexes:
```
                                                                     QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=924.55..86312.90 rows=1 width=160) (actual time=486.377..486.383 rows=1 loops=1)
```

here's teh same for the new query:

```
                                                                                    QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=1231.87..1236.91 rows=1 width=805) (actual time=19.396..19.397 rows=1 loops=1)
```

Issue #457, opened 2019-04-29
Port token replacements to Rust, improve token handling


lizziegooding:
## Phase 1

Write a replace function for the fancy-regex lib that will allow us to find groups within strings and replace them with new strings. This should be a new module within the /text processing folder of the Rust code in pt2itp. 

## Phase 2

The pt2itp JavaScript code should never import geocoder-abbreviations. Right now, we import geocoder-abbreviations in the map module and then pass it to the Rust code. All Rust code that needs the tokens should instead import that code itself. For example, the `context` struct should call geocoder-abbreviations and store all tokens itself (in the `new` call).

## Phase 3

Remove support for the `tokenless` property on features. Rather than a string, the `tokenized` property should be a vector of structs like `"tokenized": [ { "token": "main", "type": "unknown }, { "token": "st", "type": "type" }]`. this will allow us to generate `tokenless` on the fly by dropping all tokens which have a `unknown` or non-street type.

lizziegooding:
## [phase 4]

Add a final phase (this was originally implied as part of phase 3 but probably makes sense to break out separately, we'll add support for regex token replacements using the fancy-regex crate geocoder-abbreviations uses. This will require wrapping some fancy-regex types into custom types and implementing `PartialEq` and potentially other traits.

lizziegooding:
Phases 1-3 are written and awaiting additional manual review https://github.com/ingalls/pt2itp/pull/466. For this first version

*conflate will not add features where*:
- one feature is missing a road type
- one feature is missing a cardinal
- the cardinals match but are in a different position
- the road types match but are in a different position

*conflate will still add features where*:
- both features contain road types and they don't match
- both features contain cardinals and they don't match

We may ship phases 1-3 first and implement phase 4 separately as soon as we're confident in phase 3.

lizziegooding:
Posting a more complete summary of @ingalls and my work to date on conflate and token handling.

*tl;dr* the new conflate mode is correctly detecting duplicates and has dramatically reduced the number of duplicate features we're adding to Hecate. There are still known cases in which we are erroneously adding duplicate features.

## How conflate works

Conflate mode allows us to compare a new dataset with the data we already have in Hecate to ensure that we're not adding duplicate features. It performs two main actions; `create`, in which we add a new feature to Hecate and `modify`, in which we update an existing feature. Right now, the only property we're updating is alternative names (stored in the `properties.street` array).

[...]

PR #459, opened 2019-04-29
[phase 1] Support regex replacements for fancy-regex

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/457. Phase 1 of the work to port pt2itp token replacements to Rust and improve token handling. Creates a `replace` function that handles simple and capture group replacements for fancy-regex `Regex` structs. Note that like fancy-regex itself (and unlike Rust's core regex crate), `replace` does not support named capture groups.

## Next steps
- [x] get tests passing
- [x] make `replace` public, ~(maybe) implement as a method~ implement as a trait on `Regex` structs
- [x] add more tests for more replacement cases (e.g. no match, multiple numbered replacers, etc)
- [x] add better error handling if the replacement contains a named capture group
- [x] see if you can remove any code ported from the core regex crate
- [x] improve how we parse certain numbered capture groups; for example, the core regex library will interpret `$1v` to be a named capture group called `1v` rather than the numbered capture group `1` followed by the character `v`

cc @ingalls @samely @boblannon

### DIFF
--- a/native/src/text/replace.rs
+++ b/native/src/text/replace.rs
@@ -13,21 +13,25 @@ impl ReplaceAll for Regex {
 
         if rep.contains("$") {
             while input.len() > 0 {
+                // captures finds the left-most first match in a string
                 match self.captures(input)? {
                     None => {
                         new.push_str(&input);
                         break;
                     },
                     Some(m) => {
+                        // capture group 0 always corresponds to the entire match
                         let pos = (m.pos(0).unwrap().0, m.pos(0).unwrap().1);
+                        // add the string up until the beginning of the match to the output
                         new.push_str(&input[..pos.0]);
+                        // add the capture group replacement to the output
                         expand_str(&m, &rep, &mut new);
+                        // set input to the original string from the end of the match and repeat
                         input = &input[pos.1..];
                     }
                 }
             }
-        }
-        else {
+        } else {
             while input.len() > 0 {
                 match self.find(input)? {
                     None => {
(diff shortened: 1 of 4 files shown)


lizziegooding:
Noting the module I pulled some of the capture group replacement code from the core regex crate https://docs.rs/crate/regex/1.1.6/source/src/expand.rs and its implementation of `replacen` (which `replace_all` is based on) https://docs.rs/regex/1.1.6/src/regex/re_unicode.rs.html#537-591

native/src/text/replace.rs
```diff
@@ -0,0 +1,195 @@
...
+                match self.captures(input)? {
+                    None => {
+                        new.push_str(&input);
+                        break;
+                    },
+                    Some(m) => {
+                        let pos = (m.pos(0).unwrap().0, m.pos(0).unwrap().1);
+                        new.push_str(&input[..pos.0]);
+                        expand_str(&m, &rep, &mut new);
+                        input = &input[pos.1..];
```
apendleton: Looks good, but documenting for clarity here since it took me a couple of read-throughs to figure out what this is doing: it finds the left-most capture, then advances the slice to start with the end of the left-most capture and tries again.

native/src/text/replace.rs
```diff
@@ -0,0 +1,195 @@
...
+                    }
+                }
+            }
+        }
+        else {
```
apendleton: style nit: I think we generally tend towards

```
} else {
```

in our code?

PR #461, opened 2019-05-02
[phase 2] Import new geocoder-abbreviations format in Rust

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/457. Phase 2 of the work to port pt2itp token replacements to Rust and improve token handling. Changes `conflate` and `map` mode to import geocoder-abbreviations tokens in the Rust address/network importer rather than in the JavaScript wrapper. Note that tokens are converted from their original import format to match the hashmap structure pt2itp has been using. We only import non-regex tokens at the moment.

## Summary of changes
- [ ] have lib/map.js and lib/conflate.js pass an array of language codes to rust address/network importers, no longer import tokens themselves

## Next steps
This structure will change significantly in phase 3 when we start using a new structure to capture token types as well as `full`/ `canonical` variants.

cc @ingalls


native/src/text/replace.rs
```diff
@@ -185,9 +185,7 @@ mod tests {
     find!(find_cap_ref3, "$0", c!(0, 2));
     find!(find_cap_ref4, "$5", c!(5, 2));
     find!(find_cap_ref5, "$10", c!(10, 3));
-    
-    // TODO Why is this failing - investigate and re-enable
-    // find!(find_cap_ref6, "$42a");
+    find!(find_cap_ref6, "$42a", c!(42, 3));
```
lizziegooding: This wasn't failing for some reason when I originally wrote this test-- `find_cap_ref` should be able to detect a capture group given a replacement string of `$42a` where `$42` is the numbered capture group

PR #466, opened 2019-05-14
Conflate

ingalls:
Sets the stage for using a unified name matching algorithm between various modes of pt2itp.

- Rewrite `conflate` mode in rust to eliminate bugs in the JS address parsing code which have been fixed in the rust stream version
- Create `native/src/linker` as the stage for huge improvements to our matching code in conflate  - and where @lizziegooding's tokenization changes will be implemented.
- Add a ton of new unit tests around token/matching behavior
- Add `is_numeric` & `is_routish` helper functions for matching numberish street names

cc/ @lizziegooding


native/src/conflate/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+    let context = match args.context {
+        Some(context) => crate::Context::from(context),
+        None => crate::Context::new(String::from(""), None, crate::Tokens::new(HashMap::new()))
+    };
+
+    let pgaddress = pg::Address::new();
+    pgaddress.create(&conn);
+    pgaddress.input(&conn, AddrStream::new(GeoStream::new(args.in_persistent), context.clone(), args.error_persistent));
+
+    pgaddress.index(&conn);
```
boblannon: not sure if this still applies but in current master, the join against the `address` table relied in part on `address.number`:

https://github.com/ingalls/pt2itp/blob/8986ad4cf48d9f649207476f1657a62afd6bcff8/lib/conflate/compare.js#L96

If that's still true, then `pg::Address.index` should be updated such that it adds an index for `number`.  In the old code, I found a decent speedup and lower memory usage when all the concurrency threads stopped doing heap scans.

lizziegooding:
*tl;dr* the new conflate mode is correctly detecting duplicates and has dramatically reduced the number of duplicate features we're adding to Hecate. There are still known cases in which we are erroneously adding duplicate features.

## How conflate works

Conflate mode allows us to compare a new dataset with the data we already have in Hecate to ensure that we're not adding duplicate features. It performs two main actions; `create`, in which we add a new feature to Hecate and `modify`, in which we update an existing feature. Right now, the only property we're updating is alternative names (stored in the `properties.street` array).

Unlike on master, the new conflate mode will detect an existing address in Hecate in the following cases:
- one feature is missing a road type (e.g. Loveland: `1416 Santa Cruz`, Hecate: `1416 SANTA CRUZ DR`)
- one feature is missing a cardinal (e.g. Loveland: `3038 Country Club Rd E`, Hecate: `3038 COUNTRY CLUB RD`)
- cardinals match but are in a different position (e.g. Loveland: `15275 W FM 968`, Hecate `15275 FM 968 W`)
- road types match but are in a different position (e.g `Main St` vs `St Main`)
Conflate will `modify` the existing address rather than `create` a duplicate address.

The new conflate mode will still add duplicate addresses in the following cases:
- features have mismatching road types (e.g. Loveland: `2500 ALPINE RD`, Hecate: `2500 Alpine St`).
- features have mismatching cardinals (e.g. `Main St E` vs `Main St W`)
- features have differing whitespace (e.g. Loveland: `100 pine crest dr`, Hecate: `100 Pinecrest Dr`)

We are intentionally preventing conflate from detecting duplicates where cardinals and road types don't match because we want to err on the side of stricter matching and adding more points where possible. While this may add some additional noise, there are many instances where distinct proximate streets do have different cardinals or road types and should be added as new features.

Screen_Shot_2019-05-22_at_12_12_42_PM
_72nd Dr, Av, and Rd in Queens_

We should handle whitespace differences in the future. We have not yet scoped this task and it's not on our immediate roadmap.

## Stats summary

To compare the effectiveness of the new conflate mode vs master, I reran conflate using a Loveland source from Harrison County Texas. The original Loveland source contained 890 parcels.

`master`
Features added: 91
Features modified: 0

`conflate`
Features added: 25
Features modified: 32

During @samely's initial review, she found that there were only 11 new features that should actually be added of the 91 detected by conflate on `master`. We don't have a list of those exact features, but after reviewing the output of the new conflate, I found that 15 of the 25 addresses conflate identified were valid. Of the invalid features, 7 were not detected as duplicates due to known whitespace issues, and 3 appear to be invalid or non existant addresses, indicating poor quality Loveland data.

Anecdotally, the Loveland addresses we would add in this county don't appear to be particularly high value. Many of them are empty lots and only a handful appear to be commercial properties. Harrison county is quite rural and small (only 66,000 people), so this may not be the case in other counties.

PR #467, opened 2019-05-14
[phase 2] Import new geocoder-abbreviations format in Rust

lizziegooding:
Ref https://github.com/ingalls/pt2itp/pull/461

cc @ingalls


native/src/text/tokens.rs
```diff
@@ -14,30 +15,16 @@ impl Tokens {
         }
     }
 
-    pub fn generate(mut tokens: Vec<Vec<String>>) -> Self {
+    pub fn generate(tokens: Vec<String>) -> Self {
```
ingalls: Let's rename this `languages` as it no linger contains the tokens themselves

native/src/text/replace.rs
```diff
@@ -185,9 +185,7 @@ mod tests {
     find!(find_cap_ref3, "$0", c!(0, 2));
     find!(find_cap_ref4, "$5", c!(5, 2));
     find!(find_cap_ref5, "$10", c!(10, 3));
-    
-    // TODO Why is this failing - investigate and re-enable
-    // find!(find_cap_ref6, "$42a");
+    find!(find_cap_ref6, "$42a", c!(42, 3));
```
lizziegooding: This wasn't failing for some reason when I originally wrote this test-- find_cap_ref should be able to detect a capture group given a replacement string of $42a where $42 is the numbered capture group

native/src/text/tokens.rs
lizziegooding: Per chat, we decided to rename this both here and throughout pt2itp to make it more clear when we're working with language codes (which we'll use to retrieve tokens) vs when we're working with the tokens themselves

native/src/text/replace.rs
ingalls: @lizziegooding I double checked against the diff, it wasn't failing (on GH) as I hadn't set up `cargo test` in our Circle Config (my bad). It was still failing for me locally on the branch. I commented it out temporarily to get master green again when I enabled cargo test on circle

PR #468, opened 2019-05-16
[phase 3] Make tokenized an array, remove tokenless

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/457. Removes support for the `tokenless` property on features. Rather than a string, the output `tokenized` property is now an array of objects like `"tokenized": [ { "token": "main", "type": "null"}, { "token": "st", "type": "way" }]`. This will allow us to generate tokenless on the fly by dropping all tokens which have an unknown or non-street type. This work is on top of @ingalls 's changes to conflate in https://github.com/ingalls/pt2itp/pull/466.

## Summary of changes
- [ ] remove support for `tokenless` property
- [ ] make `tokenized` property an array of objects including the token type rather than a string

## Next steps
- [ ] get tests passing/incorporate any additional changes from `conflate`

cc @samely

### DIFF
--- a/native/src/text/mod.rs
+++ b/native/src/text/mod.rs
@@ -62,9 +62,8 @@ pub fn distance<T>(a: &T, b: &T) -> usize
 ///
 pub fn is_numbered(name: &Name) -> Option<String> {
     let tokens: Vec<String> = name.tokenized
-        .clone()
-        .into_iter()
-        .map(|x| x.token.to_string())
+        .iter()
+        .map(|x| x.token.to_owned())
         .collect();
 
     lazy_static! {
@@ -89,9 +88,8 @@ pub fn is_numbered(name: &Name) -> Option<String> {
 ///
 pub fn is_routish(name: &Name) -> Option<String> {
     let tokens: Vec<String> = name.tokenized
-        .clone()
-        .into_iter()
-        .map(|x| String::from(x.token))
+        .iter()
+        .map(|x| x.token.to_owned())
         .collect();
 
     lazy_static! {
@@ -201,8 +199,8 @@ pub fn syn_ca_french(name: &Name, context: &Context) -> Vec<Name> {
         && !eliminator.contains(&name.tokenized[1].token)
     {
         let tokens: Vec<String> = name.tokenized[1..name.tokenized.len()]
-            .into_iter()
-            .map(|x| x.token.to_string())
+            .iter()
+            .map(|x| x.token.to_owned())
             .collect();
         let basic = tokens.join(" ").trim().to_string();
 
--- a/native/src/text/mod.rs
+++ b/native/src/text/mod.rs
@@ -11,6 +11,7 @@ mod replace;
 
 pub use self::diacritics::diacritics;
 pub use self::tokens::Tokens;
+pub use self::tokens::Tokenized;
 
 use std::collections::HashMap;
 use regex::{Regex, RegexSet};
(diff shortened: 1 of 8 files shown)


native/src/util/linker.rs
```diff
@@ -41,25 +44,49 @@ impl LinkResult {
 /// Geometric proximity must be determined/filtered by the caller
 ///
 pub fn linker(primary: Link, mut potentials: Vec<Link>) -> Option<LinkResult> {
-    // Ensure exact matches are always returned before potential short-circuits
     for name in &primary.names.names {
-        for potential in potentials.iter() {
+        let atoks: Vec<String> = name.tokenized
```
lizziegooding: Should we keep the `atoks` and `ntoks` names? I found them somewhat confusing but wanted to use what we were already generating from `tokenized` rather than splitting again

native/src/util/linker.rs
```diff
@@ -41,25 +44,49 @@ impl LinkResult {
...
+            .clone()
+            .into_iter()
+            .map(|x| String::from(x.token))
+            .collect();
+        let tokenized = String::from(atoks.join(" ").trim());
+
+        let tokenless: Vec<String> = name.tokenized
```
lizziegooding: Doing this over and over again feels clunky-- let me know if you have thoughts, maybe moving `concat` (would rename) out of token tests https://github.com/ingalls/pt2itp/pull/468/files#diff-2c280581b529df944fe3897773c6fd68R131?

native/src/util/linker.rs
```diff
@@ -41,25 +44,49 @@ impl LinkResult {
...
+            .clone()
+            .into_iter()
+            .filter(|x| x.token_type.is_none())
```
lizziegooding: Note that this is different from what's in production in that we only drop tokens that have an associated `TokenType`, not all tokens that were tokenized. This means that for a feature named `Saint Peter Street`, in production `tokenless` would be `peter` whereas in this version of conflate it would be `st peter` (`saint` => `st` but doesn't have an associated `TokenType`). Let me know if you want to change this to replicate what we have in production in this iteration-- know we'll be building strict/loose matching next.

native/src/text/mod.rs
```diff
@@ -60,14 +61,18 @@ pub fn distance<T>(a: &T, b: &T) -> usize
 /// Is the street a numbered street: ie 1st, 2nd, 3rd etc
 ///
 pub fn is_numbered(name: &Name) -> Option<String> {
-    let tokens = name.tokenized.split(' ');
+    let tokens: Vec<String> = name.tokenized
+        .clone()
```
ingalls: We should avoid `clone` calls whenever possible, especially when this will be run every time per name. 

```
        .iter()
        .map(|x| x.token.to_string())
        .collect();
```

native/src/text/mod.rs
```diff
@@ -83,14 +88,18 @@ pub fn is_numbered(name: &Name) -> Option<String> {
 /// ie: US Route 4
 ///
 pub fn is_routish(name: &Name) -> Option<String> {
-    let tokens = name.tokenized.split(' ');
+    let tokens: Vec<String> = name.tokenized
+        .clone()
```
ingalls: Same here, let's avoid `clone()`

native/src/util/linker.rs
ingalls: @lizziegooding why don't you create a helper function on the `Name` object.

ie:
```
tokenized_string()
```

native/src/util/linker.rs
ingalls: :+1: I think this should fix a bunch of outstanding bugs

native/src/util/linker.rs
lizziegooding: Great suggestion! makes this much easier to read

PR #469, opened 2019-05-20
make number matching insensitive

samely:
`pt2itp` compare is case sensitive for numbers of type: `N453`

So if the number is `n453`, we will have duplicates while conflating.

@ingalls

### DIFF
--- a/lib/conflate/compare.js
+++ b/lib/conflate/compare.js
@@ -93,7 +93,7 @@ class Compare {
                         FROM
                             address p
                         WHERE
-                            p.number = '${new_address.properties.number}'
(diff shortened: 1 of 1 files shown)


lib/conflate/compare.js
```diff
@@ -93,7 +93,7 @@ class Compare {
                         FROM
                             address p
                         WHERE
-                            p.number = '${new_address.properties.number}'
+                            lower(p.number) = '${new_address.properties.number.toLowerCase()}'
```
samely: This is failing so I imagine it is:
`lower(p.number) = lower('${new_address.properties.number}')`

ingalls:
This code has been moved to rust and should be :tada:

PR #471, opened 2019-05-21
[phase 3] strict mode

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/457. Adds a `strict` mode to linker.rs where we're more stringent with what features are allowed to match. As a first pass, @ingalls and I decided on the following criteria:
- if way tokens exist in both the primary and the potential features, they must match (e.g. `Main Ave` cannot match `Main St` but `Main St` can match `Main`)
- if cardinal tokens exist in both the primary and the potential features, they must match (e.g. `East Main St` cannot match `West Main St` but `East Main St` can match `Main St`)
- in all other cases, we calculate the score just as we normally would

@ingalls I added a few tests for cases I'm not sure how we want to handle:
- where there are multiple cardinals/ways and only some match, e.g. `East Main` vs `North East Main`
- where one feature contains an ordinal and the other a way, e.g. `East Main` vs `Main Street`

Would love your thoughts

## Next steps
- [ ] review by @ingalls 
- [ ] merge

### DIFF
--- a/native/src/conflate/mod.rs
+++ b/native/src/conflate/mod.rs
@@ -111,7 +111,7 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
                 AND ST_DWithin(ST_SetSRID(ST_Point($2, $3), 4326), p.geom, 0.01);
         ", &[ &addr.number, &addr.geom[0], &addr.geom[1] ]).unwrap();
 
-        let mut persistents: Vec<Address> = Vec::with_capacity(rows.len());
+        let mut potentials: Vec<Address> = Vec::with_capacity(rows.len());
 
         for row in rows.iter() {
             let dist: f64 = row.get(0);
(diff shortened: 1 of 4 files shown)


native/src/conflate/mod.rs
```diff
@@ -111,7 +111,7 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
                 AND ST_DWithin(ST_SetSRID(ST_Point($2, $3), 4326), p.geom, 0.01);
         ", &[ &addr.number, &addr.geom[0], &addr.geom[1] ]).unwrap();
 
-        let mut persistents: Vec<Address> = Vec::with_capacity(rows.len());
```
ingalls: Hmm I don't think the renaming of these variables makes sense. The addresses being conflated are the potentials, they are going to be conditionally added to the database. The data being compared against is persistent - it's going to hang around no matter what.

native/src/util/linker.rs
```diff
@@ -41,19 +41,39 @@ impl LinkResult {
...
+                                if potential_name.has_type(Some(TokenType::Cardinal)) && !potential_name.tokenized.contains(tk) {
+                                    continue 'outer;
+                                }
+                            },
+                            Some(TokenType::Way) => {
+                                if potential_name.has_type(Some(TokenType::Way)) && !potential_name.tokenized.contains(tk) {
+                                    continue 'outer;
+                                }
+                            },
+                            Some(_) | None => ()
```
ingalls: This can just be abbreviated to `_`

native/src/conflate/mod.rs
lizziegooding: Good catch, I didn't read this carefully enough initially (was thinking of map mode) and wanted variable naming to match the `compare` parameter naming. What you had here originally made sense, I've it changed back.

PR #476, opened 2019-06-05
[phase 4] support regex token replacements

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/457. Phase 4 of the work to port pt2itp token replacements to Rust and improve token handling. Adds support for regex rather than whole word token replacements and uses geocoder-abbreviations' `Token` struct directly rather than mutating into the `ParsedTokens` intermediate format. See https://github.com/mapbox/geocoder-abbreviations/pull/52 for upstream updates.

## Blockers
Tests are current failing because there are tokens (so far only `Street => St` and `Saint => St` which have non-unique abbreviated forms. This means that the `TokenType` associated to that particular token is arbitrary depending on which token happens to come first in geocoder-abbreviations. Because `Saint => St` is before `Street => St`, any street name that contains the abbreviated form `St` is not detected as a `TokenType::Way`. This leads to unexpected results when we generate the `tokenless` form of a street name in the linker code, and means that in both strict and loose modes, we can't match the abbreviated form of `St` to `Street`:
https://github.com/ingalls/pt2itp/blob/bf84efa8916d6221ca8a2e54372f174918d81ebb/native/src/util/linker.rs#L315

## Next steps
- [x] see if geocoder-abbreviations contains any other tokens with non-unique abbreviated forms
- [x] figure out how to handle tokens with non-unique abbreviated forms, get tests passing
- [x] make sure we're handling these cases correctly in both the Rust and the JS code
- [ ] remove regex filtering, actually perform replacements in both Rust and JS
- [ ] revisit how you're handling lowercasing when matching inputs to tokens

cc @ingalls @samely @miccolis

### DIFF
--- a/native/Cargo.lock
+++ b/native/Cargo.lock
@@ -5,7 +5,15 @@ name = "aho-corasick"
 version = "0.6.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "aho-corasick"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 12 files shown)


lizziegooding:
Closing as stale

PR #477, opened 2019-06-10
St => Saint, Street, State

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/pull/476

Input features to pt2itp may match [geocoder-abbreviations](https://github.com/mapbox/geocoder-abbreviations) tokens with non-unique abbreviated forms tokens, for example:
- `St Peter St`
- `St Hwy 123`
where `st` could refer to `Saint`, `Street`, or `State`. This means that the TokenType associated to that particular token is arbitrary depending on which token happens to come first in geocoder-abbreviations. Because `Street => St` is after `Saint => St`, any name that contains the abbreviated form St is detected as a `TokenType::Way`. This leads to unexpected results when we generate the tokenless form of a street name in the linker code, and means that in both strict and loose modes, we can't match the abbreviated form of `St` to `Saint`:
```
// Feature A
original: St Peter St 
tokenless: peter

// Feature B
original: Saint Peter Street
tokenless: st peter
``` 
This PR attempts to guess which unabbreviated form `st` is referring to based on the number of occurrences and the presence of other `way` tokens within the original name.

## Rules for `st` detection
- if `st` is followed by `hwy` or `rte`, `st => state`, handle it like any other `state( hwy|rte)` name
- given a single `st` token in the string:
    - if a `way` type already exists that's not a `st` token, e.g. `rd`, `av`, `st` => `Saint` or `State`, `TokenType == None`. Example: `St Peter Ave`, `St Hwy 123`
    - else `st` => `Street`, `TokenType == Way`. Example: `Main St`
- given two or more `st` tokens in the string:
    - all but the last `st` token is likely `Saint` or `State`, `TokenType == None`. Example: `St Peter St`.
    - for the last `st` token
      - if a `way` type already exists that's not a `st` token, `st` => `Saint` or `State`, `TokenType == None`. Example: `St Peter Ave`, `St Hwy 123`. Example: `Saint State Ave` (probably very unlikely)
      - else `st` => `Street`, `TokenType == Way`. Example: `St Peter St`

## Next steps
- [ ] make any necessary modifications to linker.js to similarly handle `st` tokens

cc @ingalls @miccolis @samely


lizziegooding:
> make any necessary modifications to linker.js to similarly handle st tokens

It looks like this isn't necessary as linker.js draws from what's already in the database. The import step before either linker.js or linker.rs is run is the same (in Rust) and will correctly handle `st => saint, street, or state`.

PR #482, opened 2019-06-19
Map Stats

ingalls:
Add the ability to attach boundaries in which statistics should be generated about the data being processed.

Statistics will include (for each boundary)
- Accuracy Counts (`point`, `parcel`, `rooftop`)
- Total Addresses
- Total Intersections
- List of Street Names

This PR also removed `analyze` mode, a mode that we haven't used for several years and when trying to fix up tests, realizing is very very behind our current database schema. Tools like our process scripts and conflate have largely replaced this.


ingalls:
This approach turned out to be too slow - superseded by https://github.com/ingalls/pt2itp/pull/483

PR #483, opened 2019-06-20
Boundary Based Stats

ingalls:
This PR introduces the ability to give the `stats` module a list of polygonal boundary files in which individual stats will be calculated. It also drops support for the `analyze` script which has been incompatible with our database schema for quite some time. It is now superseded by our DW, hecate process & conflate based infrastructure.

### Features
- [x] `GeoStream` -> `geo` type conversion
- [x] `impl RTreeObject` on `geo` type
- [x] Load RTree with boundaries
- [x] Iterate over addresses and find where they are in the rtree
- [x] Calculate number of postcodes in boundary
- [x] Calculate accuracy tags in a given boundary
- [x] Switch `map` integration tests to use real `DC` data instead of the current limited `SG` data


native/src/stats/mod.rs
```diff
@@ -52,141 +35,219 @@ pub fn stats(mut cx: FunctionContext) -> JsResult<JsValue> {
...
-                                if elenest.is_number() || elenest.is_string() {
-                                    intsecs = intsecs + 1;
-                                }
-                            }
-                        } else if ele.is_string() {
-                            intsecs = intsecs + 1;
-                        }
-                    }
+#[derive(Serialize, Deserialize, Debug)]
+pub struct StatsBound {
```
lizziegooding: This is a nit, ok if we move these structs/methods to the beginning of the file? Makes it a little easier to navigate/diff

test/stats.test.js
```diff
@@ -106,8 +111,71 @@ test('Stats - Invalid', (t) => {
...
+                synonyms: ['101', 'US Route', 'US Route 101'],
+                custom: {
+                    postcodes: 1,
+                    accuracy: {
+                        rooftop: 1,
+                        parcel: 0,
+                        point: 1
+                    }
+                }
+            }, 'DC should have data');
```
lizziegooding: Is this DC? Does `bound` here refer to a FIPs code?

native/src/stats/mod.rs
ingalls: Done

test/stats.test.js
ingalls: The code itself is generic, in this particular case the input data is the FIPS code.

ingalls:
@lizziegooding, per the description above, it mentions removing the `analyze` mode. I'll add the DC test switch.

PR #486, opened 2019-06-28
Famous People

ingalls:
MLK/JFK are very common street names and have many common variations. We currently don't support standardizing them. 

Concretely, this query doesn't work; `7474 NE Martin Luther King Junior Blvd, Portland, OR`

However this somewhat cryptic variant works fine; `7474 NE M L King Blvd, Portland, OR`

- mlk
- mlk jr
- m l king
- m l king jr
- Martin Luther King
- Martin Luther King Junior

/cc @ingalls

### DIFF
--- a/native/src/text/mod.rs
+++ b/native/src/text/mod.rs
@@ -402,7 +402,7 @@ pub fn syn_us_famous(name: &Name, context: &Context) -> Vec<Name> {
 
         syns.push(Name::new(format!("MLK Jr{}", &strtype), -1, &context));
         syns.push(Name::new(format!("M L K Jr{}", &strtype), -1, &context));
-        syns.push(Name::new(format!("Martin Luther King Junior{}", &strtype), 1, &context));
+        syns.push(Name::new(format!("Martin Luther King Jr{}", &strtype), 1, &context));
 
     } else if MLK.is_match(name.display.as_str()) {
         let strtype: String = match MLK.captures(name.display.as_str()) {
@@ -652,7 +652,7 @@ mod tests {
         let results = vec![
             Name::new("MLK Jr", -1, &context),
             Name::new("M L K Jr", -1, &context),
-            Name::new("Martin Luther King Junior", 1, &context)
+            Name::new("Martin Luther King Jr", 1, &context)
         ];
 
         assert_eq!(syn_us_famous(&Name::new(String::from("mlk jr"), 0, &context), &context), results);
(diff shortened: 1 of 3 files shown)


native/src/text/mod.rs
```diff
@@ -368,6 +368,59 @@ pub fn syn_us_cr(name: &Name, context: &Context) -> Vec<Name> {
...
+            Some(capture) => match capture.name("type") {
+                Some(name) => name.as_str().to_string(),
+                None => String::from("")
+            },
+            None => String::from("")
+        };
+
+        syns.push(Name::new(format!("MLK Jr{}", &strtype), -1, &context));
+        syns.push(Name::new(format!("M L K Jr{}", &strtype), -1, &context));
+        syns.push(Name::new(format!("Martin Luther King Junior{}", &strtype), 1, &context));
```
lizziegooding: Do we want to add `Martin Luther King Jr` here?

native/src/text/mod.rs
```diff
@@ -368,6 +368,59 @@ pub fn syn_us_cr(name: &Name, context: &Context) -> Vec<Name> {
...
 }
 
+///
+/// Generate synonyms for names like "MLK" => "Martin Luther King"
+///
+pub fn syn_us_famous(name: &Name, context: &Context) -> Vec<Name> {
+    let mut syns: Vec<Name> = Vec::new();
+
+    lazy_static! {
+        static ref JFK: Regex = Regex::new(r"(?i)^j(ohn)?\s?f\s?k(ennedy)?(?P<type>\s.*)?$").unwrap();
```
lizziegooding: Do we want to match instances where a `.` is included? e.g. `John F. Kennedy`, `Martin Luther King Jr.`? Will these have already been stripped out?

native/src/types/name.rs
```diff
@@ -262,7 +263,7 @@ impl Name {
     ///
     /// ```
     pub fn new(display: impl ToString, priority: i8, context: &Context) -> Self {
-        let mut display = display.to_string();
+        let mut display = display.to_string().trim().to_string();
```
lizziegooding: Know we 🗣 about this briefly, do we also want to filter out empty tokens from `tokenized` while we're making these updates?

native/src/text/mod.rs
ingalls: It should be handled by our abbreviations list in carmen

Issue #494, opened 2019-07-23
Refactor display text and synonym generation


lizziegooding:
## Context

While investigating a bug where we were emitting equivalent synonyms in our final address clusters, I did a deep dive into how we process the original names we get from component address and network features and create synonyms. There's a lot here that's leftover from when we hadn't ported much of the code base to Rust or introduced more intelligent token handling. We should fix this duplication issue and make our name process pipeline simpler and more robust. See PR for next steps.

cc @miccolis @ingalls @samely

ingalls:
@lizziegooding can we expand the scope slightly to handle proper titlecasing of minors at the beginning of a text value?

Example:

```
a 10, Germany
```

should be 

```
A 10, Germany
```

My guess is that an `en` language code is potentially being passed in the `titlecase.js` module to https://github.com/mapbox/title-case, if this is the case we should adjust the behavior of how minors are capitalized to only be lowercased if not the first token in a text value.

lizziegooding:
Done in https://github.com/mapbox/pt2itp/releases/tag/25.0.0 and https://github.com/mapbox/pt2itp/releases/tag/v26.0.0

PR #495, opened 2019-07-24
Refactor display text and synonym generation

lizziegooding:
## Context

Ref https://github.com/ingalls/pt2itp/issues/494. Refactors how we generate display text and synonyms for features.

## Summary of changes
- [x] set the `priority` and `source` of generated synonyms within synonym generating functions in native/src/text/mod.rs
- [x] remove octothorpes (`#`) when creating new `Name` structs for canadian and US addresses

## Next steps
- [x] move `ext,extension,etc` processing in lib/titlecase.js to text/mod.rs. This is very similar to synonym handling we already perform in this generic text module.
- [x] don't perform a nested loop to set `freq` values when `internal_tokenized_string` is the same in lib/titlecase.js. Depending on how we end up doing the final sorting, we may not need to reset these frequency values at all
- [x] Move the `tileCase` function in lib/titlecase.js to types/name.rs so we're normalizing the display string when we create a new `Name` struct. This will make working with these strings easier and allow us to detect duplicates/further normalize earlier.
- [x] Remove duplicates based on the `display` and tokenized string values when creating a new `Names` struct in types/name.rs with a hashmap rather than in titlecase.js with lodash
- [x] ~remove `Name`s with an empty display struct when creating a new `Names` struct in types/name.rs rather than in lib/titlecase.js~ we already do this, going to remove from lib/titlecase.js
- [x] don't set `freq` and `priority` values in lib/titlecase.js-- all of this should be handled when `Name` and `Names` structs are created
- [x] (maybe) do a final names sort when we create a new `Names` struct in type/name.rs rather than in lib/titlecase.js. Need to verify we don't do any additional manipulation of names between when we create a new `Names` struct and apply lib/titlecase.js as part of post processing
- [x] deprecate https://github.com/mapbox/title-case
cc @samely @miccolis @ingalls


native/src/text/mod.rs
```diff
@@ -269,12 +258,25 @@ pub fn syn_ca_hwy(name: &Name, context: &Context) -> Vec<Name> {
...
-                if name.priority > 0 {
-                    syns.push(Name::new(format!("{} {} {}", &region_name, &hwy_type, &num), 0, &context));
-                } else {
-                    syns.push(Name::new(format!("{} {} {}", &region_name, &hwy_type, &num), 1, &context));
-                }
+                let mut syns: Vec<Name> = Vec::new();
+
+                // if the original feature has a priority of < 0, ensure the display form synonym priority isn't
+                // > the address and primary network names. Otherwise ensure it's always > the original name
+                let display_priority = if name.priority >= 0 { name.priority + 1 } else { -1 };
```
lizziegooding: This math is a bit messy but is meant to ensure we can reasonably assume the following about `priority`:
- names that come from the address cluster or are the primary name of the network have a default `priority = 0`
- names that are not the primary name of the network (synonyms that exist on the network) have a `priority = -1`
- generated synonyms that are the display name should have a higher priority than the names of original features (`priority = 1`) _unless_ the name it's derived from is lower priority (likely a non-primary network name) `priority = -1`
- generated synonyms that are _not_ the display name, but are an alternative name, should always have a lower priority than the names of original features and should always be < 0

This should effectively prevent non-primary network names (e.g. highways) from displaying primary network names (e.g. the local street name), like we do on master. 

On master, we used to set the `priority` of all non-display name generated synonyms to -1 after the fact; now we respect the priority level set in the synonym generating functions.

native/src/types/name.rs
```diff
@@ -240,6 +240,10 @@ impl Name {
         // @TODO titlecase and normalize display name here
         if context.country == String::from("US") || context.country == String::from("CA") {
             display = text::str_remove_octo(&display);
+            // penalize less desireable street names
+            if text::is_undesireable(&tokenized) {
```
lizziegooding: I moved this here so we can set the priority of less desirable street names _before_ we generate synonyms

lib/label/titlecase.js
```diff
@@ -55,51 +47,3 @@ module.exports = () => {
...
- * @param {string} text Text to be titlecased
- * @param {Array} config Object containing @mapbox/title-case configuration
- */
-function titleCase(text, config) {
-    config.minors = config.minors || [];
-    config.pre = config.pre || [];
-    config.post = config.post || [];
-    const separators = [];
-    for (let separator = wordBoundary.exec(text); separator; separator = wordBoundary.exec(text)) {
-        // preserve the characters separating words, collapsing runs of whitespace but preserving other stuff
```
lizziegooding: I did not port this portion of the titleCase function; it was not clear to me from https://github.com/ingalls/pt2itp/pull/150 what the intention of this code was beyond removing unnecessary white space. The only test that changed as a result of removing it is https://github.com/ingalls/pt2itp/blob/c1de61ce7ec9bd48723ebd08713d47516428f742/native/src/text/titlecase.rs#L112 where we used to remove the `-` before `bra` (`abra CAda -bra`=> `Abra Cada -Bra` instead of `Abra Cada Bra`). The original test case this PR fixed, `Mt. Moosilauke Hwy` still passes with the new rust implementation

native/src/text/replace.rs
```diff
@@ -147,17 +147,17 @@ mod tests {
 
     #[test]
     fn test_replace() {
-        assert_eq!(Regex::new("(?:floor|fl) #?\\d{1,3}").unwrap().replace_all("123 main st", "").unwrap(), "123 main st");
-        assert_eq!(Regex::new("(?:apartment|apt|bldg|building|rm|room|unit) #?(?:[A-Z]|\\d+|[A-Z]\\d+|\\d+[A-Z]|\\d+-\\d+[A-Z]?)").unwrap().replace_all("123 main st apt #5 washington dc", "").unwrap(), "123 main st  washington dc");
-        assert_eq!(Regex::new("(?:floor|fl) #?\\d{1,3}").unwrap().replace_all("123 main st floor 5", "").unwrap(), "123 main st ");
-        assert_eq!(Regex::new("\\d{1,3}(?:st|nd|rd|th) (?:floor|fl)").unwrap().replace_all("5th floor", "").unwrap(), "");
-        assert_eq!(Regex::new("[１1]丁目").unwrap().replace_all("1丁目 意思", "一丁目").unwrap(), "一丁目 意思");
+        assert_eq!(Regex::new(r"(?:floor|fl) #?\d{1,3}").unwrap().replace_all("123 main st", "").unwrap(), "123 main st");
```
lizziegooding: Changes these tests to use raw strings to avoid unnecessary escaping

native/src/text/titlecase.rs
```diff
@@ -0,0 +1,130 @@
+use regex::Regex;
+use crate::Context;
+
+///
+/// Titlecase input strings
+///
+
+pub fn titlecase(text: &String, context: &Context) -> String {
```
lizziegooding: Ported from lib/titlecase.js and https://github.com/mapbox/title-case

native/src/text/titlecase.rs
```diff
@@ -0,0 +1,130 @@
...
+    if context.country == String::from("US")
+    || context.country == String::from("CA") {
+        new = normalize_cardinals(&new)
+    }
+
+    new
+}
+
+pub fn capitalize(text: &str, context: &Context) -> String {
+    const MINOR_EN: [&str; 40] = [
```
lizziegooding: Note that I'm using these minor words based on the context country rather than language. We have the desired language code on the `InputContext` struct when we create the `Context` struct, but don't carry it over to `Context`. Happy to retain this property on `Context` so we can apply these minor words to english / german rather than US/Canada and germany

native/src/types/name.rs
```diff
@@ -228,16 +229,11 @@ impl Name {
...
-        let mut display = display.to_string().trim().to_string();
+        let mut display = display.to_string().replace(r#"""#, "");
+        display = titlecase(&display, &context);
 
         let tokenized = context.tokens.process(&display);
 
-        display = display
-            .replace(r#"""#, "")
-            .replace("\t", "")
-            .replace("\n", "");
```
lizziegooding: We added `\t` and `\n` normalization in https://github.com/ingalls/pt2itp/pull/407. We now replace all whitespace with a single space in titlecasing so no longer need to do this separately. We do handle `"` in titlecasing, but I left stripping it out first based on https://github.com/ingalls/pt2itp/pull/404-- @ingalls is this still something we have to worry about with postgres?

native/src/text/mod.rs
```diff
@@ -34,7 +37,7 @@ pub fn distance<T>(a: &T, b: &T) -> usize
     if v2len == 0 { return v1len; }
 
     fn min3<T: Ord>(v1: T, v2: T, v3: T) -> T{
-        std::cmp::min(v1, std::cmp::min(v2, v3))
+        min(v1, min(v2, v3))
```
ingalls: Small pushback here, generally I try not to import these globally as it means that `min` is now a reserved name unless we are using `cmp` throughout the entire mod.

native/src/types/name.rs
ingalls: @lizziegooding I'm not sure, you'd have to create some test data to see how postgres behaved

native/src/types/name.rs
```diff
@@ -262,20 +251,28 @@ impl Name {
     /// * `display` - A string containing the street name (Main St)
     ///
     /// ```
-    pub fn new(display: impl ToString, priority: i8, context: &Context) -> Self {
-        let mut display = display.to_string().trim().to_string();
+    pub fn new(display: impl ToString, mut priority: i8, source: Option<Source>, context: &Context) -> Self {
```
ingalls: Hmm do we have to require `source` here, I feel as though this makes writing code around Name less ergonomic - as can be seen with the copious `None` values that have been added to `Name::new` calls.

I've found that using the builder style function is easier as it allows source to be optionally applied where necessary while allowing single line invocation.

```
let name = Name::new().set_source()
```

lib/help.js
```diff
@@ -86,7 +86,6 @@ module.exports = function(argv) {
             console.log('   --in-address=<FILE.geojson>              geojson of address points');
             console.log('   --db="<DATABASE>"                        Name of database to connect to w/ user "postgres"');
             console.log('   --output=<FILE.geojson>                  output generated ITP lines');
-            console.log('   --label=<Label>                          Default text standardization strategy - defaults to titlecase');
```
lizziegooding: Noting that in the port to Rust I removed support for a custom titlecasing library. We could consider adding back this option if someone were to provide a library in Rust

native/src/conflate/mod.rs
```diff
@@ -218,6 +218,9 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
                 let names_new = Names::from_input(names_new, &context);
 
                 names_base.concat(names_new);
+                names_base.empty();
```
lizziegooding: I separated these out, where we used to `concat` which also deduped, this explicitly has each function perform one cleaning action

native/src/types/name.rs
lizziegooding: Per chat, requiring a source when we generate new `Name`s is necessary because if a source isn't provided, we will titlecase all display text. This means when we create new synonyms like in https://github.com/ingalls/pt2itp/pull/495/files#diff-93ff8a9065f4f7edbcc042be1f7e7e75R458, we'll convert things we've already displayed the way we want them (e.g. `MLK`) to our generic titlecasing (`Mlk`). If we were to set the source to `generated` after the fact and keep the same titlecasing, we'd have to keep track of what the original text passed in to Name::new() was and revert it, which would be messy.

test/post.text.test.js
```diff
@@ -21,59 +17,14 @@ test('Post: Text', (t) => {
...
         }
-    }, args), {
+    }), {
         properties: {
             'carmen:text': 'Main Street',
             'carmen:text_xx': 'Spring Rd'
         }
     }, 'preserve basic feature');
 
-    t.deepEquals(post({
```
lizziegooding: We now do this in rust rather than a post processing script, moved relevant test cases over to `test_names`

test/post.text.test.js
```diff
@@ -93,11 +44,15 @@ test('Post: Text', (t) => {
                 { display: '218 Haywood Rd', tokenized: [{ token: '218', token_type: null }, { token: 'haywood', token_type: null }, { token: 'rd', token_type: 'Way' }] }
             ]
         }
-    }, args), {
+    });
+
+    t.deepEquals(result, {
         properties: {
-            'carmen:text': '201 Haywood Rd,202 Haywood Rd,203 Haywood Rd,204 Haywood Rd,208 Haywood Rd,209 Haywood Rd,210 Haywood Rd,211 Haywood Rd,212 Haywood Rd,213 Haywood Rd'
+            'carmen:text': '204 Haywood Rd,201 Haywood Rd,202 Haywood Rd,203 Haywood Rd,208 Haywood Rd,209 Haywood Rd,210 Haywood Rd,211 Haywood Rd,212 Haywood Rd,213 Haywood Rd'
```
lizziegooding: The former synonym generating code used to also alphabetize names of the same priority/frequency/display_length; the current code keeps them in the input order in these cases.

test/split.intersections.test.js
```diff
@@ -66,8 +66,8 @@ test('Split: Simple Intersection', (t) => {
         t.deepEquals(res.properties, {
             'carmen:intersections': [null, null, [
                 'Germany Valley Road',
-                'Mountaineer Drive',
-                'Wv Route 33'
+                'WV Route 33',
+                'Mountaineer Drive'
```
lizziegooding: Again, this changed due to no longer sorting alphabetically

native/src/text/titlecase.rs
```diff
@@ -35,26 +38,34 @@ pub fn titlecase(text: &String, context: &Context) -> String {
...
 
-pub fn capitalize(text: &str, context: &Context) -> String {
+pub fn capitalize(text: &str, word_count: usize, context: &Context) -> String {
     const MINOR_EN: [&str; 40] = [
         "a", "an", "and", "as", "at", "but", "by", "en", "for", "from", "how", "if", "in", "neither", "nor",
         "of", "on", "only", "onto", "out", "or", "per", "so", "than", "that", "the", "to", "until", "up",
         "upon", "v", "v.", "versus", "vs", "vs.", "via", "when", "with", "without", "yet"
     ];
 
+    const MAJOR_EN: [&str; 2] = ["us", "dc"];
```
lizziegooding: We should consider adding all 2 letter state/region codes

native/src/text/titlecase.rs
```diff
@@ -123,8 +134,14 @@ mod tests {
...
+        assert_eq!(titlecase(&String::from("us hwy 1"), &context), String::from("US Hwy 1"));
+        assert_eq!(titlecase(&String::from(" -a nice road- "), &context), String::from("-A Nice Road-"));
+        assert_eq!(titlecase(&String::from("-x"), &context), String::from("-X"));
+        assert_eq!(titlecase(&String::from("x-"), &context), String::from("X-"));
+        assert_eq!(titlecase(&String::from(" *$&#()__ "), &context), String::from("*$&#()__"));
 
         let context = Context::new(String::from("de"), None, Tokens::new(HashMap::new()));
-        assert_eq!(titlecase(&String::from("Du hast recht"), &context), String::from("du Hast Recht"));
+        assert_eq!(titlecase(&String::from(" hast Du recht"), &context), String::from("Hast du Recht"));
+        assert_eq!(titlecase(&String::from("a 9, 80939 münchen, germany"), &context), String::from("A 9, 80939 München, Germany"));
```
lizziegooding: Fixes https://github.com/ingalls/pt2itp/issues/494#issuecomment-516046602

lizziegooding:
In my previous implementation, we weren't always preferencing `Name`s with a longer `display` text given the same tokenized name and equal priority. Rust's `sort` and `sort_by` methods use the [timsort](https://en.wikipedia.org/wiki/Timsort) algorithm, which expects input data to have already been sorted. This meant that we would not always compare all duplicates in the list of `Name`s and therefore not always pick the `Name` with a longer `display` text.

For example, given an input list of synonyms like:
```
NE Martin Luther King Jr Blvd, priority: 0, freq: 110
Northeast Martin Luther King Junior Boulevard, priority: 0, freq: 1
NE Martin Luther King Jr Blvd, priority: 1, freq: 1
NE MLK Blvd, priority: -1, freq: 1
NE M L K Blvd, priority: -1, freq: 1
NE Martin Luther King Blvd, priority: -1, freq: 1
NE MLK Jr Blvd, priority: -1, freq: 1
NE M L K Jr Blvd, priority: -1, freq: 1
Northeast Martin Luther King Jr Boulevard, priority: 1, freq: 1
Northeast MLK Boulevard, priority: -1, freq: 1
Northeast M L K Boulevard, priority: -1, freq: 1
Northeast Martin Luther King Boulevard, priority: -1, freq: 1
Northeast MLK Jr Boulevard, priority: -1, freq: 1
Northeast M L K Jr Boulevard, priority: -1, freq: 1
```

with no other sorting, timsort produces a deduped output of:
```
Northeast Martin Luther King Jr Boulevard, priority: 1, freq: 1
NE MLK Blvd, priority: -1, freq: 1
NE M L K Blvd, priority: -1, freq: 1
NE Martin Luther King Blvd, priority: -1, freq: 1
NE MLK Jr Blvd, priority: -1, freq: 1
NE M L K Jr Blvd, priority: -1, freq: 1
```

where we've surfaced shorter `Name`s like `NE MLK Blvd` rather than `Northeast MLK Boulevard`.

First sorting alphabetically by tokenized name fixes this problem, but means that the input order of `Name`s with the same priority is not respected. This means that synonyms with the same priority and lower frequencies can surface first.

For example, given an input like:
```
NE M L King Blvd, priority 0, freq 1480
N M L King Blvd, priority 0, freq 3
```

Produces final, normalized names like:
```
N Martin Luther King Jr Blvd
NE Martin Luther King Jr Blvd
```

because generated synonyms don't receive the same frequency values as their original names.

The solution I came up with is to sort all `Name`s by priority and frequency first, then ensure all potential duplicates are compared on deduping and input order is respected https://github.com/ingalls/pt2itp/pull/495/commits/6f9a84fa9860bb9426405cfa2c0348c725ff7ee0.

This means given an input like the first example, we get an output like:
```
Northeast Martin Luther King Jr Boulevard, priority: 1, freq: 1
Northeast MLK Boulevard, priority: -1, freq: 1
Northeast M L K Boulevard, priority: -1, freq: 1
Northeast Martin Luther King Boulevard, priority: -1, freq: 1
Northeast MLK Jr Boulevard, priority: -1, freq: 1
Northeast M L K Jr Boulevard, priority: -1, freq: 1
```

native/src/types/name.rs
```diff
@@ -174,35 +136,53 @@ impl Names {
...
+                None => {
+                    tokenized.insert(name.tokenized_string(), Dedupe::new(name, i));
+                }
             }
-
-            tokenized.insert(name.tokenized_string(), true);
-            self.names.push(name);
+        }
+        let mut names: Vec<Dedupe> = tokenized.into_iter().map(|(_,x)| x).collect();
+        names.sort_by(|a, b| a.first_index.partial_cmp(&b.first_index).unwrap());
```
mattciferri: Would it make sense to run names.dedupe(), then names.sort(), so that we can get rid of this additional sort and the Dedupe struct? From what I can tell, this `sort_by` simply returns the vec to the input order from the previous sort.

I could be missing something, but it could be nice to dedupe, then run a single sort on a smaller dataset.

lizziegooding: Totally agree, this does add in an extra sort step which isn't great. The problem with deduping before sorting is when we dedupe, we currently completely overwrite the previous `Name` struct. This means that the new `Name` may have a lower `freq` than the original `Name`, and thus will be sorted lower than it should be in the final output. 

For example, given an input of:
```
NE MLK Blvd, priority: 0, freq: 3
N MLK Blvd, priority: 0, freq: 2
Northeast MLK Blvd, priority: 0, freq: 1
```

when we dedupe, we'll drop `NE MLK Blvd` in favor of the longer version, `Northeast MLK Blvd`. If we now sort again by priority and frequency, `N MLK Blvd` will surface before `Northeast MLK Blvd`, even though there was a duplicate of the same name with a higher frequency.

We could preserve the highest frequency of any duplicate and only overwrite the `display_name` of duplicates rather than the entire `Name` struct, but this would make it challenging when debugging to figure out where a given `Name` came from if it doesn't preserve its original `freq` and `source` values.

lizziegooding:
I discovered yesterday that we still need a post processing deduping step. We have three places where duplicates can enter our pipeline: when we import address and network features, when we cluster addresses and networks, and when we cluster addresses and networks together. The first part of this process is written in Rust and can use the dedupe functionality I wrote. The second portion happens entirely in postgres, and the third in JavaScript. Eventually, we should dedupe within the clustering addresses ↔ network code, but until that’s migrated to Rust,  I [created a wrapper around the dedupe](https://github.com/mapbox/pt2itp/pull/495/commits/88024ce9986c52168c44dd1dc2ac8c52ce33307f) functionality I already wrote that will do a final dedupe at the end.

native/src/map/mod.rs
```diff
@@ -208,3 +208,40 @@ pub fn intersections(mut cx: FunctionContext) -> JsResult<JsBoolean> {
...
+        None => Vec::new(),
+        Some(arg) => {
+            if arg.is_a::<JsUndefined>() || arg.is_a::<JsNull>() {
+                Vec::new()
+            } else {
+                let arg_val = cx.argument::<JsValue>(0)?;
+                neon_serde::from_value(&mut cx, arg_val)?
+            }
+        }
+    };
```
ingalls: If names array is 0 we should short circuit

native/src/types/name.rs
```diff
@@ -211,7 +189,13 @@ impl Names {
             } else if a.priority < b.priority {
                 std::cmp::Ordering::Greater
             } else {
```
ingalls: Small format nitpick, can we just turn this into a single if/else instead of a nestled if/else inside the larger if/else block

lizziegooding: Per chat, this should remain nested so we only sort on frequency if priorities are equal

lizziegooding:
CI is currently failing due to a [coveralls outage](https://status.coveralls.io/) that's preventing us from uploading coverage stats. Tests are all passing so going to go ahead and merge.
Screen Shot 2019-08-12 at 12 05 11 PM

PR #499, opened 2019-08-09
Only run splitter once

lizziegooding:
## Context

As part of debugging why we're occasionally seeing duplicate address clusters in output, adds a condition to ensure the `splitter()` function is only ever run once.

cc @miccolis @samely @ingalls

### DIFF
--- a/lib/map.js
+++ b/lib/map.js
@@ -179,11 +179,8 @@ function main(argv, cb) {
                 child.stderr.pipe(process.stderr);
 
                 child.on('exit', () => {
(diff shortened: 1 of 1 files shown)


lizziegooding:
Uupdated to only run the `orphans` func, which creates orphan address clusters, once.

lizziegooding:
CI is currently failing due to a [coveralls outage](https://status.coveralls.io/) that's preventing us from uploading coverage stats. Tests are all passing so going to go ahead and merge.
Screen Shot 2019-08-12 at 12 05 11 PM

PR #502, opened 2019-08-13
Split words on graphemes not bytes during titlecasing

lizziegooding:
## Context

In the previous implementation of the Rust port of titlecasing, we were indexing into strings using bytes rather than chars or graphemes. This PR changes our titlecasing implementation to capitalize words based on [graphemes](https://en.wikipedia.org/wiki/Grapheme) instead. Issues with indexing into strings in Rust are clearly documented in https://doc.rust-lang.org/book/ch08-02-strings.html#indexing-into-strings (which I've definitely read, but conveniently forgot). I've used the [unicode_segmentation](https://docs.rs/unicode-segmentation/1.3.0/unicode_segmentation/index.html) crate for grapheme detection, which was ported from the now deprecated core Rust implementation.

## Next steps
- [ ] review
- [ ] merge
- [ ] release

cc @miccolis @ingalls @samely @mattciferri

### DIFF
--- a/native/Cargo.lock
+++ b/native/Cargo.lock
@@ -575,6 +575,7 @@ dependencies = [
  "serde 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.85 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.37 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


lizziegooding:
CI is still failing due to a [coveralls outage](https://status.coveralls.io/) that's preventing us from uploading coverage stats. Tests are all passing so going to go ahead and merge.
Screen Shot 2019-08-13 at 2 44 12 PM

PR #508, opened 2019-08-22
Panic given non-deterministic primary name priority values on networks

lizziegooding:
## Context

We currently omit networks and write to an error file networks whose first two names have the same priority value. We do this to ensure on sort and insertion into the database, we can ensure a deterministic primary name and don't experience wiggle between builds. These types of errors are usually caused by issues in our synonym generation code or our input network geojson. This PR changes this test to a hard fail.

## Summary of changes
- [ ] panic if the first two names of a given network feature have the same priority
- [ ] removes repeated code in net.rs

## Next steps
- [ ] run test builds with current branch, see if there are instances where it's valid for networks to have the same priority

cc @samely @mattciferri @miccolis @ingalls


native/src/text/mod.rs
```diff
@@ -150,9 +150,9 @@ pub fn is_undesireable(tokenized: &Vec<Tokenized>) -> bool {
     let subs = vec![
             String::from("ext"),
             String::from("connector"),
-            String::from("br"),
             String::from("unit"),
             String::from("apt"),
+            String::from("apts"),
```
lizziegooding: Added this to account for an additional abbreviation we support https://github.com/mapbox/geocoder-abbreviations/blob/26442ab0ce9f3f6e4b09b1f1cb795478420d2d73/tokens/en.json#L421-L428

native/src/text/mod.rs
```diff
@@ -150,9 +150,9 @@ pub fn is_undesireable(tokenized: &Vec<Tokenized>) -> bool {
     let subs = vec![
             String::from("ext"),
             String::from("connector"),
-            String::from("br"),
```
lizziegooding: Removed because it was identifying too many false positives, for example `Piney Branch Trail Northwest` which is a perfectly valid street name that shouldn't be penalized

PR #512, opened 2019-08-29
Tune behavior of raiseHigh & dropLow

ingalls:
`0` addresses can be valid address numbers in many place in the US, however they are more commonly the representation of null values in a strictly typed GIS system. Often we don't have a good metric for removing the address points associated with them.

This can occasionally cause problems as they can potentially affect the interpolation range of the output feature. This PR omits them from being considered by the `raiseHigh` / `dropLow` code. 

The thought here is that if they are valid there should be a `2`/`4`/etc value near them that will be selected and cause the `dropLow` code to cover the range. If the `0` is not valid, it will let the much higher initial number be used instead, generating a more sensible range.

This PR also affects the behavior of the `diff` function. Before it would generate the largest nth that it could base on the range. These ranges become very expensive when they are overly large, and as such as now brought down by a factor of 10^-1 when they exceed 1000.


lib/map/split.js
```diff
@@ -296,13 +313,16 @@ class Split {
...
     /**
      * Join addresses from a single MultiPoint collection into an array of [ Address, Corresponding Network Segments, Numbers ]
+     *
      * @param {Object} network GeoJSON FeatureCollection of network segments
      * @param {Array} coords Coordinates for address points
-     * @param {Array} props Corresponding address properties - parallel to coordinates array
+     * @param {AddressNumber[]} props Corresponding address properties - parallel to coordinates array
      * @param {Array} intersections Corresponding intersection points
-     * @return {Array} Return array of matched segments
+     * @param {Number|undefined} prevDelta
```
ingalls: Can we add a description here?

lib/map/interpolize.js
```diff
@@ -295,20 +302,20 @@ function interpolize(splits, argv = {}) {
...
-                    result.lparity = 'O';
+                    result.parityl = 'O';
                 } else {
-                    result.lparity = 'O';
+                    result.parityl = 'O';
                 }
             }
 
-            if (result.lparity === 'E') {
+            if (result.parityl === 'E') {
```
samely: I don't think we should change the `lparity` to `parityl`, since the e other variable names since to match the patern `side-name`

miccolis: Sadly this naming strangeness is "correct" because it is what the Tiger data spec prescribes; see https://pelias.io/milestones/interpolation/existing_standards.html ...`parityl` is correct.

PR #513, opened 2019-09-04
Address sequence smoothing

samely:
## Context
The output of `interpolize.js` contains cases where the ranges overlap between each other, especially if there is a bad address data.

Example:

lparity | lfromhn | ltohn | rparity | rfromhn | rtohn
-- | -- | -- | -- | -- | --
E | 10530 | 10962 | O | 10531 | 10965
E | 10968 | 1150 | O | 10975 | 11023
E | 11102 | 11248 | O | 11101 | 11251

We can see that 1150 is not the right interpolation range and overlap with other ranges. We expect a number in the 10K-11K range instead of 1500.

@miccolis @ingalls @lizziegooding @mattciferri


lib/map/interpolize.js
```diff
@@ -461,6 +340,233 @@ function join(itps, argv) {
...
+            if (result.lstart && result.lstart.number % 2 !== 1) result.lstart.number++;
+            if (result.lend && result.lend.number % 2 !== 1) result.lend.number++;
+        }
+    }
+
+    return result;
+}
+
+/**
+ * Check if there is overlapping between the ranges. If so, recalculates ranges.
```
ingalls: Can we have a better description of what this function does? For the new reader, what does "recalculate ranges" mean

lib/map/interpolize.js
```diff
@@ -185,204 +177,91 @@ function interpolize(splits, argv = {}) {
...
-            result.rstart.geometry.properties.right = true;
-        }
-        if (result.lstart) {
-            result.lstart.geometry.properties.start = true;
-            result.lstart.geometry.properties.left = true;
-        }
-        if (result.rend) {
-            result.rend.geometry.properties.end = true;
-            result.rend.geometry.properties.right = true;
+        if (feature) {
```
ingalls: Why does this need to be conditional? The `genFeat` function should always return an object, no?

lib/map/interpolize.js
```diff
@@ -461,6 +340,233 @@ function join(itps, argv) {
...
+        if (!ltoInRange) itp.rangeOptions.exclude.push(['lend', lto]);
+        if (!rfromInRange) itp.rangeOptions.exclude.push(['rstart', rfrom]);
+        if (!rtoInRange) itp.rangeOptions.exclude.push(['rend', rto]);
+
+        const range = calculateRanges(itp.rangeOptions);
+        let resFeat = genFeat(itp.genFeatOptions.split, range, itp.genFeatOptions.argv, itp.genFeatOptions.intersections);
+
+        resFeat = itp.genFeatOptions.argv.debug ? setDebug(range, resFeat) : resFeat;
+
+        if (resFeat) itp = resFeat;
```
ingalls: One again, does this need to be conditional? `genFeat` should always return

test/split.split.test.js
```diff
@@ -231,3 +232,80 @@ test('Split: Ensure cluster#break roads are split (Washington St)', (t) => {
         t.end();
     });
 });
+
+test('todo', (t) => {
```
ingalls: This should be named something other than todo

lib/map/interpolize.js
samely: well, it was there already, I just moved code, but yeah, `genFeat()` doesn't return null at any time

lib/map/interpolize.js
samely: same here I was following the existent conditional.

test/split.split.test.js
samely: This is @miccolis change from sort-fixup but will change it if he doesn't when merging his PR.

miccolis: Oh, sorry. Let's fix in the other PR

PR #518, opened 2019-09-13
Improve moving average calculation

miccolis:
Back on https://github.com/mapbox/pt2itp/pull/353 a moving average calculation was added that looked a little suspect to me. This PR modifies that calculation so that it's calculating a true moving average and make the calculation much faster. On a test of 1000 elements run 10k times the old code too 142,000 ms & the new code takes 100ms

I do have a lingering question; did the old code generate something that wasn't a moving average on purpose? @ingalls ... any idea?


lib/map/cluster.js
```diff
@@ -301,7 +293,31 @@ class Cluster {
                 floating_sin = avg[i] > 0;
             }
         }
+    }
 
+    /**
+     * Generate moving average
+     *
+     * @param {Array[]} distDelta
```
samely: Can you add a brief description per each parameter?

lib/map/cluster.js
```diff
@@ -301,7 +293,31 @@ class Cluster {
...
+     * @param {Number} avg_window
+     * @return {Number[]}
+     */
+    static movingAverage(distDelta, avg_window) {
+        const len = distDelta.length;
+        const avg = [];
+        let v = distDelta[0][1];
```
samely: not a stopping reason but, `v, w` as variable names are not very friendly.

miccolis: haha, I got in to `C` variable name mode. `v` is value and `w` is for window. I can expand them.

miccolis:
I did some testing data for DC, RI, VT, MT and CA ...and basically this has no impact on processing time. Despite that we should consider still merging this as it fixes the average calculation and adds tests for it.

test/cluster.detectPeaks.test.js
```diff
@@ -105,3 +105,87 @@ test('Assert multiple peaks detection', (t) => {
...
+    t.deepEqual(avg, [1, 1, 1, -1, -1, 1, 1, 1, 1, 1]);
+
+    vals = [1, 1, 1, 1, 1, 1, -1, -1, -1, -1, -1, -1, -1, 1, 1, 1, 1, 1, 1, 1].map((v, i) => [i, v]);
+    avg = Cluster.movingAverage(vals, 2);
+    t.deepEqual(avg,[1, 1, 1, 1, 1, 1, 0, -1, -1, -1, -1, -1, -1, 0, 1, 1, 1, 1, 1, 1]);
+
+    t.end();
+});
+
+test.skip('Moving Average - performance', (t) => {
```
miccolis: Just realized I left this in, it was helpful for validating an issue I saw, but not useful long term. Gonna yank it out.

PR #519, opened 2019-09-19
More smoother

miccolis:
Building on the work that @samely did to smooth out bumps in interpolation ranges, this branch is an attempt to add more tests and address some more cases. I've limited work here far to things I discovered while adding tests and groking the codebase. I feel like we do need to make bigger changes, but would like to have tests in place first.

Current changes;

* Change method name `generateInterpolationRanges` (plural) to `generateInterpolationRange` (singular) to reflect that it acts on a single range at a time.
* Move distance and parity calculation into new `calculateInterpolationParams` method and add some light tests.
* Refactor range adjustment out of `checkInterpolationRanges` so that it's just is actually just checking ranges
* Change median distance from line calculation, so that it uses the 50% percentile distance, not some random point in the middle position of the array.
* Some comment & code adjustment for readability
* Change the `isInRange` check to be more vigorous
* Added unit tests for `generateInterpolationRange`, `checkInterpolationRanges` and `isInRange`

Questions:

1. ~There are some minor test fixture changes that could use review.~
2. ~There was explicit `sequence` checking in `checkInterpolationRanges` that I didn't quite follow and have commented out right now. I could use some background here.~
3. ~We're explicity `delete`ing some `itp` attributes & I don't follow why. Commented them out here.~


lib/map/interpolize.js
```diff
@@ -535,62 +556,75 @@ function generateInterpolationRanges(options) {
...
+ * @param {number} i defaults to 0
+ * @return {null|number|string}
+ */
+function propAtIdx(feature, property, i = 0) {
+    return feature.properties[property] ? feature.properties[property][i] : null;
+}
 
+/**
+ * Determine if an number is between its neighbors
+ * @param {number} previous
```
ingalls: Can we have descriptions on these parameters

miccolis: Honestly not sure what to write here. Between the function description and the param type and name, it's all kinda there.

test/interpolize.test.js
```diff
@@ -697,13 +697,13 @@ test('Interpolize - sequence', (t) => {
...
     ], 'ltohn is as expected');
     t.deepEquals(res.properties['carmen:rfromhn'], [
-        [11001, 3789, 3963, 4153, 4409, null, 4721, 4815, 5207, 5709, 6001, 6415, null, 7211, 7713, 7911, 8403, 8721, 9133, 9513, 10117, 10531, 10975, 11101, 11743, 12315, 12701, 13201, 13603, 14021, 14535, 15019], null
+        [11001, 3809, 3963, 4153, 4409, null, 4721, 4815, 5207, 5709, 6001, 6415, null, 7211, 7713, 7911, 8403, 8721, 9133, 9513, 10117, 10531, 10975, 11101, 11743, 12315, 12701, 13201, 13603, 14021, 14535, 15019], null
     ], 'rfromhn is as expected');
     t.deepEquals(res.properties['carmen:rtohn'], [
-        [11739, 3961, 4123, 4345, 4613, null, 4731, 5171, 5705, 5965, 6319, 6609, null, 7519, 7713, 8311, 8715, 9029, 9511, 9815, 10529, 10965, 11023, 11251, 12125, 12619, 12923, 13441, 14015, 14527, 15015, 15215], null
+        [11739, 3955, 4123, 4345, 4613, null, 4731, 5171, 5705, 5965, 6319, 6609, null, 7519, 7713, 8311, 8715, 9029, 9511, 9815, 10529, 10965, 11023, 11251, 12125, 12619, 12923, 13441, 14015, 14527, 15015, 15215], null
```
lizziegooding: Do you have a sense for why some of these ranges had small address number changes where the order or overlapping issues still exist?

test/interpolize.test.js
```diff
@@ -930,3 +930,348 @@ test('Interpolize - genFeat', (t) => {
...
+        },
+        address: {
+            type: 'Feature',
+            properties: {},
+            geometry: {
+                type: 'MultiPoint',
+                coordinates: [
+                    [-64.27004098892212, 44.54781775558832],
+                    [-64.26878571510315, 44.548093013403566],
+                    [-64.26933288574217, 44.55552448238052],
```
lizziegooding: For these tests, can you comment specifically which address number/coordinates is the outlier? In this case they are address number outliers, but they wouldn't necessarily be-- makes it easier reading the test to know what to look for

test/split.split.test.js
```diff
@@ -42,9 +42,9 @@ test('Split: Dulaney Valley Rd, MD', (t) => {
...
-        t.deepEquals(res.properties['carmen:rfromhn'], [[1, 838, 903, 1021, null, 1024, 1300, 1320, 1508, 1802, 2100, 2012, 2300, 2300, 2118, 2204, 2300, 12101, 12301, null, 12559, 12605, 12721, 12901, 13015, 13201, null, null, null, null, null, null],null], 'carmen:rfromhn should be stable');
-        t.deepEquals(res.properties['carmen:rtohn'], [[903, 714, 1021, 1021, null, 956, 1200, 1300, 1400, 1510, 1806, 2002, 2296, 2310, 2106, 2118, 2300, 12101, 12301, null, 12605, 12615, 12901, 13007, 13115, 14001, null, null, null, null, null, null], null], 'carmen:rtohn should be stable');
+        t.deepEquals(res.properties['carmen:ltohn'], [[null, null, null, null, 1001, null, 1213, 1301, 1411, null, 1806, null, 2101, 2101, null, 2203, null, null, null, 12450, null, 12720, 12900, 13006, 13114, 13216, 14000, null, null, null, null, null], null], 'carmen:ltohn should be stable');
+        t.deepEquals(res.properties['carmen:rfromhn'], [[1, 838, 903, 1021, null, 1024, 1300, 1320, 1508, 1802, 1908, 2012, 2292, 2300, 2110, 2204, 2300, 12101, 12301, null, 12559, 12605, 12721, 12901, 13015, 13201, null, null, null, null, null, null], null], 'carmen:rfromhn should be stable');
```
lizziegooding: 🎉 seeing some great fixes!

lib/map/interpolize.js
```diff
@@ -403,13 +448,13 @@ function generateInterpolationRanges(options) {
...
         if (options.overlap) {
-            options.overlap.forEach((range) => {
-                if (range && range[0] && range[1] && result[range[0]] && result[range[0]].number === range[1]) {
-                    result[range[0]] = null;
-                }
-            });
+            for (const range of options.overlap) {
+                const k = range[0];
+                const number = range[1];
+                if (result[k] && result[k].number === number) result[k] = null;
```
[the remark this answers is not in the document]
miccolis: `if result[k]` is defined, yes it will have a number... if by some bug it doesn't we'll compare `undefined` to our number and it'll still be ok`

lib/map/interpolize.js
```diff
@@ -31,186 +35,72 @@ function interpolize(splits, argv = {}) {
...
-                'distFromEnd':  turf.distance(turf.point(split.network.geometry.coordinates[split.network.geometry.coordinates.length - 1]), addr),
-                'geometry': addr,
-                'number': parseInt(split.number[address_it].number),
-                'output': split.number[address_it].output,
-                'props': split.number[address_it].props,
-                'side': null
-            };
-
-            distsFromLine.push(res.distFromLine);
+            // TODO why do we generate this feature if we don't have interpolation data?
```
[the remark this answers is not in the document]
miccolis: Agreed, I'll remove this

lib/map/interpolize.js
```diff
@@ -535,62 +582,83 @@ function generateInterpolationRanges(options) {
...
-
-        const lfromInRange = isInRange(status.prelfrom, lfrom, status.nextlfrom);
-        const ltoInRange = isInRange(status.prelto, lto, status.nextlto);
-        const rfromInRange = isInRange(status.prerfrom, rfrom, status.nextrfrom);
-        const rtoInRange = isInRange(status.prerto, rto, status.nextrto);
-
-        itp.rangeOptions.overlap = [];
+    // Current implementation is very simplistic and designed only to catch extremely wrong points
+    // with the expectation that they occur rarely. If the previous range was invalid we make no
+    // assumptions that it can be use to detect errors in the current range.
```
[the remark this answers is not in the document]
miccolis: I'll try

test/interpolize.test.js
miccolis: So this range actually got wider because of a fix to how the comparison is done.

PR #525, opened 2019-10-17
Linker: Cardinaled Address => Non Cardinaled Street

ingalls:
```
| Umpqua St
|
| . (N Umpqua St)
| . (N Unpqua St)
| S Umpqua St
| . (S Umpqua St)

Cardinaled addresses should match a proximal non-cardinaled street
before they are matched againts a further away mismatched cardinal street

In the above example (N Umpsqua St) should always match Umpqua St
and not S Umpqua St (previous behavior)
```

cc/ @mapbox/search

### DIFF
--- a/native/src/text/replace.rs
+++ b/native/src/text/replace.rs
@@ -136,7 +136,7 @@ fn find_cap_ref<T: ?Sized + AsRef<[u8]>>(
 /// Modified to only support numbered capture groups
 fn is_valid_cap(b: &u8) -> bool {
     match *b {
-        b'0' ... b'9' => true,
+        b'0' ..= b'9' => true,
         _ => false,
     }
 }
--- a/native/src/util/linker.rs
+++ b/native/src/util/linker.rs
@@ -64,7 +64,6 @@ impl LinkResult {
 /// reasons.
 ///
 pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<LinkResult> {
-    println!("{:?}", primary);
     for name in &primary.names.names {
         let tokenized = name.tokenized_string();
         let tokenless = name.tokenless_string();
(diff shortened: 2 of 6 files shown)


native/src/text/replace.rs
```diff
@@ -136,7 +136,7 @@ fn find_cap_ref<T: ?Sized + AsRef<[u8]>>(
 /// Modified to only support numbered capture groups
 fn is_valid_cap(b: &u8) -> bool {
     match *b {
-        b'0' ... b'9' => true,
+        b'0' ..= b'9' => true,
```
miccolis: Not sure I follow the implications of this - what is this doing?

native/src/util/linker.rs
```diff
@@ -305,7 +340,37 @@ mod tests {
             ];
             assert_eq!(linker(a, b, false), Some(LinkResult::new(14, 100.0)));
         }
+        */
```
miccolis: Why is all this commented out?

native/src/util/linker.rs
```diff
@@ -689,6 +754,6 @@ mod tests {
             let b = vec![Link::new(2, &b_name)];
             assert_eq!(linker(a, b, true), None);
         }
-
+        */
```
miccolis: Why is all this commented out?

native/src/text/replace.rs
ingalls: This was a beta feature that has now stabilized. The stabilized syntax is `..=` instead of the beta syntax of `...`

native/src/util/linker.rs
ingalls: Good catch, I miss my `tape.only`

miccolis: Not sure it would have worked in this case - but you can run tests by name with like `cargo test [testname]`

ingalls: @miccolis yeah it doesn't work to unit tests

PR #526, opened 2019-10-22
Add address-styles post script

mattciferri:
### Context

To better support a variety of address styles, we can take advantage of the address_style prop passed though for each address and generate an address_styles prop for each cluster. This will denote which address styles are contained in the cluster, and does not contain duplicates.

### Summary of changes

- [ ] Adds a post script for building the address_styles array for each cluster.
- [ ] Flips the post script on for map mode after props.

cc @ingalls @samely @lizziegooding @miccolis

### DIFF
--- a/lib/map/post.js
+++ b/lib/map/post.js
@@ -10,7 +10,8 @@ const defaults = [ // Mandatory Steps
     require('../post/sort').post,
     require('../post/centre').post,
     require('../post/internal').post,
-    require('../post/props').post
+    require('../post/props').post,
+    require('../post/address-styles').post
 ];
 
 /**
--- a/lib/post/address-styles.js
+++ b/lib/post/address-styles.js
@@ -1,0 +1,31 @@
+'use strict';
+
+/**
+ * Exposes a post function to add a property denoting which address styles are in a feature
+ * @param {Object} feat     GeoJSON Feature to apply address styles prop to
+ * @return {Object}         Output GeoJSON feature to write to output
+ */
+function post(feat) {
+    if (!feat) return feat;
+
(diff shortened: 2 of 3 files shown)


lib/post/address-styles.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    if (feat.properties['carmen:addressprops'] && feat.properties['carmen:addressprops']['carmen:address_style']) {
+        Object.values(feat.properties['carmen:addressprops']['carmen:address_style'])
+            .forEach((val) => {
+                styles.add(val);
+            });
+    }
+
+    if (styles.size > 0) feat.properties['carmen:address_styles'] = Array.from(styles);
```
mattciferri: @ingalls Is it same to assume that if no styles are set on the feature, then we shouldn't apply the `carmen:address styles` prop?

lib/map/post.js
```diff
@@ -10,7 +10,8 @@ const defaults = [ // Mandatory Steps
     require('../post/sort').post,
     require('../post/centre').post,
     require('../post/internal').post,
-    require('../post/props').post
+    require('../post/props').post,
+    require('../post/address-styles').post
```
mattciferri: @ingalls Let me know if you think this should always run, or if it should be optional.

ingalls: Nope, this should always be run

lib/post/address-styles.js
ingalls: Yes, if there are no styles we should omit the `carmen:address_styles` prop.

Let's make sure we update the scope document to say that if it is omitted, it just contains standard addresses. But I'm sure they will assume that.

PR #528, opened 2019-10-28
Only consider modifying addresses where output==true in conflate

lizziegooding:
## Context

As part of conflate mode, we consider modifying persistent addresses if a new address contains names the persistent address does not. We do not, however, filter out persistent addresses with `output` set to `false`. We want to add addresses with `output: false` downstream when running conflate so we don't add new addresses that we've already updated or deleted during conflate. Without adding this check, it's possible for multiple versions of the same feature to match to the link id and for conflate to attempt to modify multiple features with the same id (which `panic`s)

## Summary of changes 
- [ ] only consider modifying addresses where output==true in conflate

## Next steps
- [ ] add tests
- [ ] review
- [ ] merge
- [ ] release

cc @mattciferri @ingalls @miccolis @samely

### DIFF
--- a/native/src/conflate/mod.rs
+++ b/native/src/conflate/mod.rs
@@ -139,7 +139,9 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
         match compare(&addr, &mut persistents) {
             Some(link) => {
                 let mut link: Vec<&mut Address> = persistents.iter_mut().filter(|persistent| {
-                    if link == persistent.id.unwrap() {
+                    // only consider modifying persistent addresses that match a new address
+                    // and have output set to true
+                    if link == persistent.id.unwrap() && persistent.output == true {
                         true
                     } else {
                         false
--- a/native/src/conflate/mod.rs
+++ b/native/src/conflate/mod.rs
@@ -9,7 +9,6 @@ use neon::prelude::*;
 
 use crate::{
     Names,
-    Source,
     Address,
     hecate,
     util::linker,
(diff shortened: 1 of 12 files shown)


native/src/conflate/mod.rs
```diff
@@ -52,12 +51,19 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
             if arg.is_a::<JsUndefined>() || arg.is_a::<JsNull>() {
                 ConflateArgs::new()
             } else {
-                let arg_val = cx.argument::<JsValue>(0)?;
+                let arg_obj = cx.argument::<JsObject>(0)?;
+
+                let db = arg_obj.get(&mut cx, "db")?;
```
lizziegooding: This sets the default database to `conflate` if that arg isn't provided

native/src/conflate/mod.rs
```diff
@@ -194,47 +208,50 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
...
-                names_base.empty();
-                names_base.sort();
-                names_base.dedupe();
+            }
+            names_base.empty();
+            names_base.sort();
+            names_base.dedupe();
+            let mut names_final = Vec::with_capacity(names_base.names.len());
+            for name in names_base.names {
+                names_final.push(InputName::from(name));
```
lizziegooding: Previously we didn't standardize output names to be hecate compatible if a single persistent address matches multiple new addresses and should be updated

native/src/conflate/mod.rs
```diff
@@ -126,47 +134,53 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
...
-                            new_names.push(InputName::from(name));
+                match pmatches.len() {
+                    // if all matches have output set to false, don't modify
+                    0 => continue,
+                    1 => {
+                        let paddr = pmatches.pop().unwrap();
+                        // if the new address has names the persistent address does not, modify persistent
+                        if paddr.names.has_diff(&addr.names) {
+                            // preference new names over persistent names
+                            let mut combined_names = addr.names;
```
lizziegooding: We used to only keep persistent names that matched to new names; now we keep all persistent names (including ones that didn't match to new names) and all new names, preferencing new names

native/src/conflate/mod.rs
```diff
@@ -194,47 +208,50 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
     for mut modified in modifieds {
         let modified_obj = modified.as_object_mut().unwrap();
         let mut props = modified_obj.remove(&String::from("properties")).unwrap();
-
-        if props.as_array().unwrap().len() == 1 {
-            let props = props.as_array_mut().unwrap().pop().unwrap();
+        let props_arr = props.as_array_mut().unwrap();
```
miccolis: Nice consolidation.

native/src/conflate/mod.rs
miccolis: Not quite sure I follow here; so previously we'd pass on a object with `display`, `priority`,  `source`, `tokenized` and `freq`?

native/src/conflate/mod.rs
```diff
@@ -126,47 +134,53 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
...
+                    1 => {
+                        let paddr = pmatches.pop().unwrap();
+                        // if the new address has names the persistent address does not, modify persistent
+                        if paddr.names.has_diff(&addr.names) {
+                            // preference new names over persistent names
+                            let mut combined_names = addr.names;
+                            combined_names.concat(paddr.names.clone());
+                            combined_names.empty();
+                            combined_names.sort();
+                            combined_names.dedupe();
```
miccolis: Is it really necessary to do this dedupe and empty checking here? Can we just concat and sort? ...just seems like we're iterating over these names a lot. Maybe something for later.

native/src/conflate/mod.rs
miccolis: I could have just missed this, but why are we defaulting to preferring new names? ...this is just carrying forward the old behavior?

...also, this is where we're going to have trouble if we've improved the quality of street names, right? If we've corrected "4" to "4th Street" and then re-conflate the source with "4" it would come back. It seems to me that we may still have access to the older versions of the feature and may be able to reference those to avoid re-adding them. Some restructing may be required for that.

native/src/conflate/mod.rs
lizziegooding: Yes exactly, the output feature would contain all of those properties. `tokenized` and `freq` are not Hecate compatible. Also this format is different than the format we use for creating a new address or modifying an address when there's only a single modification to a persistent address

native/src/conflate/mod.rs
lizziegooding: We did not used to prefer new names-- I changed this because we'll usually want newer or more up to date names to take precedence. Names are still sorted and deduped so the best possible version of duplicate names is preserved.

Re having access to older versions of the feature-- our downstream implementation is to have past versions of a feature included in persistent addresses with `output == false`. This is technically a "history" but not one pt2itp knows anything about. It does mean that in the case you mentioned above, `4` street would be included in persistent addresses and if the new feature was still called `4` rather than `4th Street`, we would detect a match and not add `4` to the output.

native/src/conflate/mod.rs
lizziegooding: Per chat, I think we should keep this. While the persistent data should have already been run through this process, new addresses have not. This ensures the data we're adding does not contain duplicate names, has the best (longest untokenized) version of those names, and doesn't include empty names.

PR #529, opened 2019-10-28
Add interpolation ranges for the raised and dropped numbers at the ends

samely:
## Context
The raise and lower condition at the end of the ranges make possible to have a wider range but also cause an inaccurate location for the interpolated result at the edges.

This is an example of lowered/raised numbers, where `819/800` were lowered to `1/0` and `1123/1126`
were raised to `2001/2000`

parityl | lfromhn | ltohn | parityr | rfromhn | rtohn
-- | -- | -- | -- | -- | --
"O" | 1| 837 | "E" | 0 | 844
"O" | 845 | 2001  | "E" | 900 | 2000 

This PR attempts to fix this issue by adding new interpolated ranges with the raised/lowered numbers, so the ranges will look like:

parityl | lfromhn | ltohn | parityr | rfromhn | rtohn
-- | -- | -- | -- | -- | --
"O" | 1  | 819 | "E" | 0  | 800
"O" | 819| 837 | "E" | 800 | 844
"O" | 845 | 1123 | "E" | 900 | 1126
"O" | 1123 | 2001 | "E" | 1126 | 2000

@lizziegooding @ingalls @miccolis @mattciferri


lib/map/interpolize.js
```diff
@@ -101,50 +102,226 @@ function interpolize(splits, argv = {}) {
...
+ * @param  {Object}  itp    Interpolated addresses
+ * @param  {Object}  range  Extended range
+ * @param  {Number}  number Address number to drop/raise
+ * @param  {Boolean} isPrev If the range is at the beggining of the segment or not.
+ * @return {Object}         Extended range with extended geometry
+ */
+function buildExtRangeGeometry(itp, range, number, isPrev) {
+    if (!itp || !itp.properties || !itp.properties['carmen:addressnumber']) return;
+
+    const addressnumbers = itp.properties['carmen:addressnumber'][1].map((num) => parseInt(num));
```
ingalls: I'm confused at your intent here for generating the geometry. Why do we need to take points in account?

When we talked before I thought we were going to just shave the ends of the linestring
```
Original
-----------------------
--                   --
Start                End
```

Can you document how exactly this range extension is working?

test/split.split.test.js
```diff
@@ -256,56 +256,16 @@ test('Split - Generate sorted ranges', (t) => {
...
-            '626-610',
-            '654-634',
-            '660-660',
-            '690-690',
-            '710-704',
-            '746-740',
-            '760-1000'
-        ], lranges, 'generated expected lranges');
+        t.deepEqual(lranges,[
+            'null-null', '1-25', '53-73', '105-117', '133-137', '205-231', 'null-null', '301-313', '355-355', 'null-null', '443-469', '481-489', '501-511', '531-547', '600-592', '626-610', '654-634', '660-660', '690-690', '710-704', '746-740', '760-760', '760-1000'
```
ingalls: Can we leave these on their own lines? Intent being for easier diffing

test/split.split.test.js
```diff
@@ -256,56 +256,16 @@ test('Split - Generate sorted ranges', (t) => {
...
-            '635-611',
-            '653-637',
-            '669-661',
-            '695-685',
-            '711-701',
-            '749-731',
-            '1001-765'
-        ], rranges, 'generated expected rranges');
+        t.deepEqual(rranges,
+            ['0-4', '4-26', '48-70', '104-126', '150-150', '212-212', '230-230', '302-314', 'null-null', '400-412', '442-470', '480-480', '500-526', '530-548', 'null-null', '635-611', '653-637', '669-661', '695-685', '711-701', '749-731', '789-765', '1001-789']
```
ingalls: Same here: individual lines if possible

lib/map/interpolize.js
```diff
@@ -101,50 +104,136 @@ function interpolize(splits, argv = {}) {
...
 
-        const rightEnd = lastNonNull(itps, 'carmen:rfromhn');
-        if (rightEnd >= 0) {
-            if (itps[rightEnd].properties['carmen:rfromhn'][0] > itps[rightEnd].properties['carmen:rtohn'][0]) {
-                itps[rightEnd].properties['carmen:rfromhn'][0] = raiseHigh(itps[rightEnd].properties['carmen:rfromhn'][0], d);
-            } else {
-                itps[rightEnd].properties['carmen:rtohn'][0] = raiseHigh(itps[rightEnd].properties['carmen:rtohn'][0], d);
-            }
-        }
+    const isSameIdx = leftSide && rightSide && leftSide.idx === rightSide.idx;
```
miccolis: Aren't this check and the one above partially duplicative because `leftSide` and `rightSide` both have to evaluation to true at this point?

lib/map/interpolize.js
```diff
@@ -101,50 +104,136 @@ function interpolize(splits, argv = {}) {
...
+
+    if (isSameIdx || leftSide) {
+        const spliceIdx = isStart ? leftSide.idx : leftSide.idx + 1;
+        let range = duplicateFeature(itps[leftSide.idx]);
+
+        range = assingItpProperties(range, leftSide, isSameIdx ? rightSide : null);
```
miccolis: Why does it only matter if we're looking at the same index on this side?

lib/map/interpolize.js
```diff
@@ -101,50 +104,136 @@ function interpolize(splits, argv = {}) {
...
+
+    const itp = itps[sideOpt.idx];
+    const fromhn = itp.properties[fromProp][0];
+    const tohn = itp.properties[toProp][0];
+
+    const isAscendent = (isStart && fromhn < tohn) || (!isStart && fromhn > tohn);
+
+    // Do not raise/lower number if the segment's length is less than 10 meters
+    const segment = getEdgeSegment(itp, isAscendent ? fromhn : tohn);
+    if (!segment || turf.length(segment, { units: 'kilometers' }) <= 0.01) return;
```
miccolis: kilometers is the default so you don't need to specify it. Also, i think you can use the simpler 'distance' calculation for such a short length. 

Also, it looks like your passing though the pointOnLine info on the rangeOpt so maybe you don't need to calculate this at all because you already have it.

lib/map/interpolize.js
```diff
@@ -101,50 +104,136 @@ function interpolize(splits, argv = {}) {
...
+    itp.geometry.geometries[1].coordinates = [];
+    return itp;
+}
+
+/**
+ * Creates a new range based on the first/last range
+ * @param  {Object} feat Address feature
+ * @return {Object}      Address feature
+ */
+function duplicateFeature(feat) {
```
miccolis: Can or should we target this better? I'm a JSON parse|stringify clone is pretty expensive.

lib/map/interpolize.js
samely: `!leftSide && !rightSide` ensure both are not null, but still one of them could be `null`.

lib/map/interpolize.js
samely: This condition will work in two cases:
- When both have the same index or 
- When `leftSide` is not null
Only if both sides have the same index I need to pass  `rightSide` otherwise null (because I could still have `rightSide` but with a diff index)

lib/map/interpolize.js
samely: Yeah, this makes sense. I can try using the info in `rangeOpt`

lib/map/interpolize.js
samely: I see I can actually remove this function and create it in `assingItpProperties`

lib/map/interpolize.js
samely: Ok, I've checked this. I think it is not safe to use just distance from the point to one od the edges because not all the road are flat we could have some curved ones.

PR #533, opened 2019-11-04
Refactor address evaluation during interpolation

miccolis:
Very much a work in progress and subject to all manner or changes... this branch attempts to centralize review of address numbers so that we can better split clusters with address number resets, ignore numbers that shouldn't be used for interpolation, more cleanly create ranges across intersections and more intelligently buffer out cluster ends


lib/map/cluster.js
```diff
@@ -69,240 +59,271 @@ class Cluster {
                     intersections: seg.intersections
                 };
             });
-        } else if (!segDup.some((dup) => { return dup; })) {
-            if (this.warn) {
-                this.warn.write(`WARN: detected unhandled ClusterDup on network_cluster.id: ${id}\n`);
-            }
+        } else if (this.warn) {
+            this.warn.write(`WARN: detected unhandled ClusterDup on network_cluster.id: ${id}\n`);
```
ingalls: This doesn't exist anymore and will never be reached

lib/map/interpolize.js
```diff
@@ -1,13 +1,14 @@
 'use strict';
 
-module.exports = interpolize;
+module.exports = {
+    interpolize
```
ingalls: Just curious why this change is necessary?

lib/map/interpolize.js
```diff
@@ -23,17 +24,19 @@ const misc = new Misc();
...
  * @param {boolean} argv.debug Output debug information on features
- * @return {Object} GeoJSON Feature GeometryCollection of combined PT and calculated ITP
+ * @return {Object} GeoJSON MultiLineString of calculated ITP
  */
 function interpolize(splits, argv = {}) {
-    const prevDelta = splits.prevDelta;
-    const nextDelta = splits.nextDelta;
+    let prevDelta;
+    let nextDelta;
+    if (!Array.isArray(splits) && Array.isArray(splits.segs)) {
```
ingalls: This is just for testing right? We'll remove this in the final code?

lib/map/interpolize.js
```diff
@@ -100,14 +93,16 @@ function interpolize(splits, argv = {}) {
...
         // Otherwise if we have addresses > than the 1 million threshold, skip the drop/raise
         || limits.max > 1000000
-    ) return join(itps, argv);
+    ) return genInterpolationFeature(itps, argv.debug);
+
 
     const d = diff(limits.max, limits.min);
 
+
     if (prevDelta === undefined) itps = addExtendedRange(itps, d, true);
```
ingalls: This check is always going to be undefined as prevDelta is never set

lib/map/interpolize.js
```diff
@@ -44,54 +47,44 @@ function interpolize(splits, argv = {}) {
...
-            feature = argv.debug ? setDebug(range, feature) : feature;
-
-            // Note: rangeOptions and genFeatOptions will be present on some iterpolation ranges
-            // but not others depending on whether or not they are re-calculated below. We should
-            // tidy this up at some point and not expose this state.
-            feature.rangeOptions = rangeOptions;
-            feature.selectedRange = range;
-            feature.genFeatOptions = { split, argv, intersections: split.intersections };
+    for (const split of splits) {
+        if (!split.addressPoints || split.addressPoints.length === 0) continue;
```
ingalls: What happens to splits with no address points?

lib/map/split.js
```diff
@@ -2,34 +2,54 @@
...
- * @property {AddressNumber[]} number
- * @property {Array} network - LineString GeoJSON features
- * @property {Array} intersections - MultiPoint GeoJSON features
- * @property {Number|undefined} prevDelta Address number difference from previous cluster.
- * @property {Number|undefined} nextDelta Address number difference to next cluster
+ * @typedef {Object} AddressNumber  TODO rename
+ *
+ * @property {Number} id
+ * @property {Boolean} outlier - Address will not be used to calculate interpolation ranges if true. TODO implement
+ * @property {Boolean} output - Address will be in are in the final point output in true.
```
ingalls: Grammar

lib/map/split.js
```diff
@@ -315,91 +368,233 @@ class Split {
...
+        for (let i = 0; i < dists.length; i++) {
+            const start = i === 0 ? 0 :  dists[i - 1];
+            const end = dists[i];
+
+            // Ignore 0 length segments, ie; from intersections at start or end of segment
+            if (start === end)  continue;
+
+            // Review the addresses and attempt to match them to this segment. If we've already
+            // considered all addresses, or don't have any points on a segment it won't be present
+            // in the final output. We may want to reconsider this.
```
ingalls: We return a street level result in a reverse geocode if there isn't a proximal address point or interpolatable point. This is rare in urban areas but happens frequently in rural areas. We should not drop any of the road network unless it is erroneous (ie very tiny segment, ???, etc)

lib/map/split.js
```diff
@@ -2,34 +2,54 @@
...
 const Post = require('./post');
 
-const interpolize = require('./interpolize');
+const interpolize = require('./interpolize').interpolize;
 const Explode = require('./explode');
 const Cluster = require('./cluster');
 const Queue = require('d3-queue').queue;
 const turf = require('@turf/turf');
 const pg = require('pg');
+const { dedupe_syn } = require('../../native/index.node');
```
ingalls: The reason this was a post script was so that we don't have to replicate the functionality in multiple places. Moving the code here will break intersections for orphaned networks

miccolis: Ok, I'm having trouble finding a test for this behavior. Can you point me to a test for orphan intersections?

lib/map/split.js
miccolis: Got it. Should be resolved now.

lib/map/interpolize.js
miccolis: Yea, need to look at reverting this. At one point I'd exposed multiple methods from this file. Presently it's not the case.

lib/map/interpolize.js
miccolis: Yep

lib/map/interpolize.js
miccolis: We'll need to preserve the functionality some way - hopefully something more elegance than this - so yea, plan is to remove these.

lib/map/interpolize.js
miccolis: As written... nothing, but as per your other comments, that needs to change so that we can generate intersection only clusters.

lib/map/interpolize.js
```diff
@@ -556,14 +449,14 @@ function generateInterpolationRange(options) {
...
-            }
-        }
+        // // Null out overlapping ranges so they are populated on the next iteration.
+        // if (options.overlap) {
+        //    for (const range of options.overlap) {
+        //        const k = range[0];
+        //        const number = range[1];
+        //        if (result[k] && result[k].number === number) result[k] = null;
+        //    }
+        // }
```
miccolis: This needs to be removed.

lib/map/orphan.js
```diff
@@ -229,35 +230,39 @@ class Orphan {
...
-                                'carmen:ltohn': [[]],
-                                'carmen:rtohn': [[]]
-                            },
-                            geometry: {
-                                type: 'GeometryCollection',
-                                geometries: [
-                                    row.geom
-                                ]
-                            }
+                        // TODO check opts.args.intersections
```
miccolis: Remove this comment.

lib/map/split.js
```diff
@@ -257,53 +289,96 @@ class Split {
...
+                    m.push([a[i].location, deltaNum > 0 ? 1 : -1]);
+                }
+                return m;
+            }, []);
+
+            const breaks = [];
+            Cluster.detectCliffs(breaks, cluster.addressPoints, distDelta);
+            Cluster.detectPeaks(breaks, cluster.addressPoints, distDelta);
+
+            acc.push(...Split.splitCluster(cluster, breaks).map(Split.detectOutliers));
```
miccolis: Should outlier detection happen when we iterate over clusters below?

PR #537, opened 2019-11-18
Add outlier detection

mattciferri:
## Context

As per https://github.com/mapbox/pt2itp/issues/536, we want to improve the way we handle outliers in address data. This PR adds a step in split process that will detect and flag outliers before reaching itp range generation.

### Summary of changes

- [x] Adds tests for covering various outlier scenarios
- [x] Adds a detectOutliers function to the Cluster module
- [x] Integrates the outlier detection step into the pipeline before cluster splitting

cc @miccolis @lizziegooding @ingalls


package.json
```diff
@@ -53,6 +53,7 @@
     "pg-cursor": "^2.0.0",
     "pluralize": "^8.0.0",
     "progress": "^2.0.0",
+    "regression": "^2.0.1",
```
miccolis: I'm required to ask... doesn't anything change if we just use the regression method in simple-statistics?

lib/map/split.js
```diff
@@ -596,6 +597,56 @@ class Split {
...
+            const addrY = parseInt(addr.number);
+            const lineY = slope * addr.location + intercept;
+
+            addr.dist = lineY - addrY;
+            return addr;
+        });
+
+        const sortedRange = distFromFitLine
+            .map((v) => v.dist)
+            .sort((a, b) => a - b);
```
miccolis: Looks like we're iterating over the same data twice here. Is it worth consolidating these?

lib/map/split.js
```diff
@@ -596,6 +597,56 @@ class Split {
...
     }
+
+    /**
+     * Detect outliers in a cluster
+     *
+     * @param {AddressLocations[]} addressLocations
+     * @returns {AddressLocations[]}
+     */
+    static detectOutliers(addressLocations) {
+        if (addressLocations.length < 5) return addressLocations;
```
miccolis: Could you add a comment about why 5 is the minimum?

lib/map/split.js
```diff
@@ -596,6 +597,56 @@ class Split {
...
+            if (upper > lower && (addr.dist > upper || addr.dist < lower) ||
+                upper < lower && (addr.dist < upper || addr.dist > lower)) {
+                addr.ignore = true;
+            }
+            return addr;
+        });
+        return response;
+    }
+
+    static range(input) {
```
miccolis: This needs a js doc right?

package.json
mattciferri: The reason I brought in `regression` is that is contains useful methods for polynomial regressions, which I'm playing around with for peak detection. The simple-statistics lib only contains a linear regression method. If we ultimately end up just leveraging a linear regression, then I'll use only simple-statistics.

To answer to the question is I'm not sure, but I'm happy to do a comparison.

lib/map/split.js
mattciferri: Sort ended up not being necessary in the first place, as the quantile calculation does a quickselect.

lib/map/split.js
```diff
@@ -338,7 +342,8 @@ class Split {
             Cluster.detectCliffs(breaks, cluster.addressPoints, distDelta);
             Cluster.detectPeaks(breaks, cluster.addressPoints, distDelta);
 
-            acc.push(...Split.splitCluster(cluster, breaks));
+            acc.push(...Split.splitCluster(cluster, breaks).map(Split.detectOutliers));
```
miccolis: @mattciferri I think I know the answer, but... why we do this after cliff/peak detection and splitting. What happens if we do this first?

lib/map/split.js
```diff
@@ -605,6 +610,165 @@ class Split {
...
+    /**
+     * Merge address point, interpolation range and intersection feature into a single
+     * GeometryCollection
+     *
+     * @param {Object} itps GeoJSON MultiLineString
+     * @param {Object} pts GeoJSON MultiPoint
+     * @param {Object|undefined} intersections GeoJSON MultiPoint, Intersections can be disabled and will undefined here.
+     * @return {Object} GeoJSON GeometryCollection
+     */
+    static mergeFeatures(itps, pts, intersections) {
```
miccolis: HRM why is this here? is the PR out of date?

lib/map/split.js
```diff
@@ -605,6 +610,165 @@ class Split {
...
+        const upperTenth = ss.quantile(locationRange, 0.9);
+
+        const fitLine = regression.linear(coords);
+        const [slope, intercept] = fitLine.equation;
+
+        // return if the correlation isn't strong enough
+        if (fitLine.r2 < 0.3) return cluster;
+
+        const distsFromFitLine = addressPoints.map((addr) => {
+            const addrY = parseInt(addr.props.number);
```
miccolis: Does a condition exist where this could assign `NaN` as the value of addrY? ...if that happens what are the implications?

test/cluster.detectCliffs.test.js
```diff
@@ -119,3 +119,1959 @@ test('Assert multiple cliffs detection', (t) => {
...
+
+
+// cliff detection
+
+//                        o
+//             *              *
+//         *
+//     *              *
+// *             *
+test.skip('handle cliffs with outliers', (t) => {
```
miccolis: Whats the intent of all these skipped tests?

test/split.detectOutliers.test.js
```diff
@@ -0,0 +1,620 @@
...
+    t.deepEqual(ignored(res), []); // no outliers
+    t.end();
+});
+
+// standard outlier detection cases
+
+//             *
+//    3x*
+// *
+// TODO: add more points, this is being skipped
```
miccolis: Want to add another point or two to this test? ...and the next one?

test/split.detectOutliers.test.js
```diff
@@ -0,0 +1,620 @@
...
+// line of best fit edge cases
+
+//          *       *
+//
+//
+//                     *
+//              *               *
+//        *                     *
+// *
+// TODO: some outliers can throw off best fit line when there aren't many points
```
miccolis: Is this still an issue? ...same question for the same TODO in the below tests

lib/map/split.js
mattciferri: The idea is to identify cliffs/peaks first (using moving average-like strategies to reduce outlier noise) to split down to the closest we can get to a single linear representation. From there we run the linear regression and have more confidence in removing outliers from it.

lib/map/split.js
mattciferri: Whoops, I rebased yesterday and thought this was valid so I kept it in. I'll remove it.

lib/map/split.js
mattciferri: Good catch, the current behavior is that it simply won't flag any outliers. Tweaking things so that outliers can still be detected.

test/cluster.detectCliffs.test.js
mattciferri: These will be enabled with cliff/peak detection in the next PR.

test/split.detectOutliers.test.js
mattciferri: Ooh yes good catch

test/split.detectOutliers.test.js
mattciferri: It is but it's rare. There's not much we can do about it with the current implementation so I included these tests to capture it. I'll write up a potential solution that we can take on as a follow up - the level of effort is a bit higher.

lib/map/split.js
miccolis: Not sure, I follow... as I understand it if the parse fails we'll assign `NaN` to `add.dist`. But then below, when building the `flaggedOutliers` if we fail to parse the address number, we just let is pass through, but I'd expect to have it marked with `ignore = true`

mattciferri: Blah, wasn't thinking that it'd be good to flag them as ignore. I guess I assumed something downstream would handle that.

PR #539, opened 2019-11-27
Prepare persistent addresses before conflate

lizziegooding:
## Context

Follow up to https://github.com/mapbox/pt2itp/pull/528. From that PR:

> As part of conflate mode, we consider modifying persistent addresses if a new address contains names the persistent address does not. **We do not, however, filter out persistent addresses with `output` set to `false`. We want to add addresses with `output: false` downstream when running conflate so we don't add new addresses that we've already updated or deleted during conflate.** Without adding this check, it's possible for multiple versions of the same feature to match to the link id and for conflate to attempt to modify multiple features with the same id (which `panic`s)

This PR changes this assumption so pt2itp is responsible for setting past versions of features to `output=false` and inversing their `id`s. This allows us to match new addresses to past versions of features and not add them while also ensuring we don't attempt to modify past features.

I've chosen to perform this operation post indexing to hopefully speed up the update. @ingalls let me know if you have thoughts re performance on this; can also do some testing down stream.

cc @ingalls @miccolis @mattciferri


native/src/pg/address.rs
```diff
@@ -125,3 +125,35 @@ impl InputTable for Address {
...
+// Prepare address to be run through conflate by setting output=false and id to its inverse
+// for all past versions of a feature.
+// This ensures that past features are not modified but will match addresses being conflated.
+pub fn pre_conflate(conn: &Connection) {
+    conn.execute(r#"
+        DROP TABLE IF EXISTS address_id_to_version;
+    "#, &[]).unwrap();
+
+    conn.execute(r#"
+    CREATE TABLE IF NOT EXISTS address_id_to_version as
```
ingalls: Small formatting nitpick, can we indent these so they are separate from their function invocation?

lizziegooding:
Tests are failing because I pushed the dev tag associated with the last commit twice-- going to merge anyway

PR #541, opened 2019-12-02
Cluster splitting

mattciferri:
Reworks the way we split clusters based on the AddressCluster data structure.

### Next Steps 
- [x] Add additional test cases
- [ ] Add safeguard to prevent overlapping itp ranges


lib/map/split.js
```diff
@@ -411,11 +418,56 @@ class Split {
...
+        for (let i = 0; i < breaks.length; i++) {
+            const isFirst = breaks[i - 1] === undefined;
+
+            const breakStartLocation = isFirst ? 0 : midpoint(cluster.addressPoints, breaks[i - 1]);
+            const breakEndLocation = midpoint(cluster.addressPoints, breaks[i]);
+
+            const slice = turf.lineSliceAlong(cluster.network, breakStartLocation, breakEndLocation);
+
+            const intersectionPoints = cluster.intersectionPoints.filter((intsec) =>
+                turf.pointOnLine(slice, intsec.coords).properties.dist === 0
```
miccolis: Shouldn't we be able to compare the existing placement to the break location?

lib/map/split.js
```diff
@@ -411,11 +418,56 @@ class Split {
...
-        return [cluster];
+    static splitCluster(cluster, breaks) {
+        if (!breaks || breaks.length === 0) return [cluster];
+
+        const res = [];
+
+        breaks.sort((a, b) => a - b);
+
+        for (let i = 0; i < breaks.length; i++) {
+            const isFirst = breaks[i - 1] === undefined;
```
miccolis: This is kinda a weird check, could it just test if `i == 0`?

lib/map/split.js
```diff
@@ -411,11 +418,56 @@ class Split {
...
+            const addrEnd = breaks[i] + 1;
+
+            res.push({
+                network: slice,
+                addressPoints: cluster.addressPoints.slice(addrStart, addrEnd),
+                intersectionPoints
+            });
+
+            // add remaining segment
+            if (i === breaks.length - 1) {
```
miccolis: nitpick, if you're gonna have a first check at the top of the loop, why not a last one too?

lib/map/split.js
```diff
@@ -411,11 +418,56 @@ class Split {
...
+                const finalBreakStart = breakEndLocation;
+                const finalBreakEnd = turf.length(cluster.network);
+
+                const finalSlice = turf.lineSliceAlong(cluster.network, finalBreakStart, finalBreakEnd);
+
+                const finalIntersectionPoints = cluster.intersectionPoints.filter((intsec) =>
```
miccolis: At this point we'll have always reviewed some items in this array, can we skip some cycles here and jump right in where the remaining ones are?

lib/map/split.js
mattciferri: Hah, yeah I think the old implementation was convincing me that I had to check `i - 1` for undefined for some reason.

lib/map/split.js
mattciferri: Yes, ended up tracking an index through the intersection array and adding the points that fit between the start and end locations.

lib/map/split.js
```diff
@@ -407,15 +414,67 @@ class Split {
...
+            if (isLast) {
+                const finalBreakStart = breakEndLocation;
+                const finalBreakEnd = turf.length(cluster.network);
+
+                const finalIntersectionPoints = [];
+                while (intsecIdx < cluster.intersectionPoints.length) {
+                    if (intsec && intsec.location >= finalBreakStart && intsec.location <= finalBreakEnd) {
+                        finalIntersectionPoints.push(intsec);
+                    } else {
+                        console.error('not ok - orphan intersection found in cluster split');
```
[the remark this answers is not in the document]
mattciferri: Not carrying forward behavior as intersections weren't really accounted for in this part of the code. That said, this log statement was mostly for testing purposes to ensure intersections weren't being dropped. Now that this is all covered by tests, I can remove it.

mattciferri:
@miccolis Sure I can remove the old Cluster.break code as part of this PR. I've figured out an approach to end-of-segment case, just need to incorporate it in a way that doesn't mangle the function too much.

PR #544, opened 2019-12-12
Improve cliff/peak detection

mattciferri:
### Context

Our current method for identifying address resets (referred to as cliffs/peaks) has a few limitations, and doesn't account for all cases that we'd like to account for. By revisiting these functions we can improve the way we divide up clusters, which can result in better interpolation ranges.

### Summary of changes

- Reworks the detectCliffs and detectPeaks functions to better identify cluster breaks.
- Adds several tests to account for various real world scenarios.

### Next steps

- [ ] Perform a test run on actual data.

cc @ingalls @miccolis


lib/map/cluster.js
```diff
@@ -1,72 +1,124 @@
 'use strict';
 
+
```
miccolis: Need to remove this extra line

lib/map/interpolize.js
```diff
@@ -65,6 +65,20 @@ function interpolize(splits, argv = {}) {
         }
     }
 
+    if (itps.length === 0) {
+        const coordinates = splits.map((v) => v.network.geometry.coordinates)
+        const properties = {
+            'carmen:rangetype': 'tiger',
```
miccolis: This is assigned in mergeFeatures. Not need to include it here.

PR #547, opened 2020-04-02
Add consensus module

mattciferri:
Adds consensus module for deriving consensus agreement amongst address data.

cc @ilissablech @mapbox-danny


native/src/consensus/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+                'source', p.source,
+                'props', p.props,
+                'geom', ST_AsGeoJSON(p.geom)::TEXT
+            )
+        FROM
+            address p
+        WHERE
+            p.number = $1
+            AND p.source = $4
+            AND ST_DWithin(ST_SetSRID(ST_Point($2, $3), 4326), p.geom, 0.01);
```
bookest, line 114: [nit] Would it be possible to pull the magic numbers (`4326` and `0.01`) out to constants?

native/src/consensus/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+                    match pmatches.len() {
+                        0 => continue,
+                        1 => {
+                            let paddr = pmatches.pop();
+                            let coords = match paddr {
+                                Some(point) => {
+                                    Some((point.geom[0], point.geom[1]))
+                                },
+                                None => None
+                            };
```
bookest, line 150: [nit] IMHO `Option.map` is a clearer way to express this:
```suggestion
                            let coords = paddr.map(|p| (p.geom[0], p.geom[1]);
```

native/src/consensus/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+                            // update source_map with current match for source
+                            source_map.entry(source.to_string()).and_modify(|e| *e = coords);
+                        },
+                        _ => panic!("Duplicate IDs are not allowed in input data")
```
bookest, line 154: Would it be reasonable to include which IDs are duplicated in the error message here? If it's something that can be expressed in a reasonable way including them could help shorten the debug cycle here.

native/src/consensus/mod.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+    let addr_link = linker::Link::new(0, &addr.names);
+    let potential_links: Vec<linker::Link> = potentials.iter().map(|potential| {
+        linker::Link::new(potential.id.unwrap(), &potential.names)
+    }).collect();
+
+    match linker::linker(addr_link, potential_links, true) {
+        Some(link) => Some(link.id),
+        None => None
+    }
```
bookest, line 189: [nit] Another case where `Option.map` might be a better fit
```suggestion
    linker::linker(addr_link, potential_links, true).map(|link| link.id);
````

File: native/src/consensus/mod.rs [hunk 5]
```diff
@@ -0,0 +1,190 @@
...
+use super::pg;
+use super::pg::{Table, InputTable};
+
+#[derive(Serialize, Deserialize, Debug)]
+struct ConsensusArgs {
+    db: String,
+    context: Option<super::types::InputContext>,
+    threshold: Option<u32>,
+    sources: Option<Vec<String>>,
+    query_points: Option<String>,
```
bookest, line 24: [nit] Should `sources` and `query_points` be `Option`s if they're required? 

I know that ultimately these values are coming from JavaScript, so the checks here are pushed to runtime for a lot of use cases, but I still think there's value in leveraging the type system to make invalid states unrepresentable on the rust side. Perhaps by moving the runtime checks to the JavaScript side to keep nicer errors while still enforcing the required type level constraints on the rust side.

lib/help.js
```diff
@@ -160,6 +160,24 @@ module.exports = function(argv) {
...
+            console.log('usage: index.js consensus [<FILE>, <FILE>, ...] [--db <DATABASE>] [--threshold <INTEGER>]');
+            console.log('                    [--languages=<CODE, ...>] [--country=<CODE> ] [--region <CODE>]');
+            console.log('                    [--error-sources <FILE>] [--error-query-points <FILE>]');
+            console.log('');
+            console.log('[options]:');
+            console.log('   <FILE.geojson>                            line-delimited geojson of address features');
+            console.log('   --db="<DATABASE>"                         Name of database to connect to w/ user "postgres"');
+            console.log('   --threshold="<INTEGER>"                   [optional] Maximum allowed dissimilarity between points. Defaults to 25.');
+            console.log('   --error-sources=<FILE>                    [optional] File to log source address import errors');
+            console.log('   --error-query-points=<FILE>               [optional] File to log query point import errors');
```
bookest, line 176: The `error-sources` and `error-query-points` are marked as required in native/src/consensus/mod.rs but optional here.

mattciferri: Where are they marked as required in `native/src/consensus/mod.rs`?

native/src/consensus/agreement.rs
```diff
@@ -0,0 +1,236 @@
...
+            .map(|x| labels[x])
+            .collect();
+        for source in modal_cluster {
+            self.results.entry(String::from(source)).and_modify(|e| e.agreement_count += 1);
+        }
+    }
+}
+
+fn haversine((lon1, lat1): (f64, f64), (lon2, lat2): (f64, f64)) -> f64 {
+    const EARTH_RADIUS: f64 = 6371.0; // meters
```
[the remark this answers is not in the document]
mattciferri: Oof, yes.

lib/util/context.js
```diff
@@ -24,7 +24,7 @@ class Context {
             if (typeof args.languages === 'string') {
                 this.languages = args.languages.split(',');
             } else if (Array.isArray(args.languages)) {
-                this.langauges = args.languages;
+                this.languages = args.languages;
```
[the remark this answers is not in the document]
mattciferri: Seriously lol

native/src/consensus/agreement.rs
```diff
@@ -0,0 +1,236 @@
...
+
+        let mut source_hit_count = 0;
+        let mut coordinates = vec![];
+        let mut labels = vec![];
+        for (source, coord) in source_map.iter() {
+            if let Some(point) = coord {
+                source_hit_count += 1;
+                labels.push(source);
+                coordinates.push(point);
+                self.results.entry(String::from(source)).and_modify(|e| e.hit_count += 1);
```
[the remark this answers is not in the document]
mattciferri: I take the keys of the `source_map` and pass them directly into `Agreement::new`, so there theoretically shouldn't be any odd cases, but I agree that it's best to not keep two things floating around with no real checks that they contain the same values. I removed the sources arg and populated `results` dynamically in the `process_points` method and it cut down on some unnecessary code :+1:

native/src/consensus/mod.rs
mattciferri: Yep, good call. This section was largely lifted from another module, and I think the intent of keeping the checks within Rust was to provide clean messaging in the case where the module is called directly, rather than through the command line wrapper (`index.js`). I agree that we should just leverage the type system and have Rust/ throw an error if these parameters aren't provided.

native/src/consensus/mod.rs
mattciferri: Wow nice, didn't know this existed.

PR #548, opened 2020-04-15
Pass in the country to conflate

mapbox-danny:
Check if the country is United states when requiring tokenless matches to have identical first characters

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,8 +100,9 @@ if (require.main === module) {
             break;
         }
         case ('conflate'): {
+
             const conflate_arg = require('minimist')(process.argv, Context.args({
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -122,7 +123,8 @@ if (require.main === module) {
                 languages: conflate_arg.languages,
                 hecate: conflate_arg.hecate,
                 context: new Context(conflate_arg).as_json(),
-                db: conflate_arg.db
+                db: conflate_arg.db,
+                country: conflate_arg.country
```
mattciferri, line 127: The country param is actually built into the context, so no need to add an arg here :+1:

PR #549, opened 2020-04-21
German regex parsing

ilissablech:
Add in functionality to apply regex patters to [native/src/text/tokens.rs](https://github.com/mapbox/pt2itp/blob/de_strasse/native/src/text/tokens.rs). 

For Germany only, for streets ending in "strasse", "str, or "straße" we will split these endings off into its own word. And then apply the token matching. For example: "banhofstrasse" => "banhof" => "bhf".

In testing a small cluster of addresses this change brought address_orphans down from 1183 to 212 & network_orphans down from 1764 to 801.

### DIFF
--- a/native/Cargo.lock
+++ b/native/Cargo.lock
@@ -311,7 +311,7 @@ dependencies = [
 [[package]]
 name = "geocoder-abbreviations"
 version = "0.1.0"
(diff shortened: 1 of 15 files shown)


ilissablech:
Closing in lieu of a cleaner PR https://github.com/mapbox/pt2itp/pull/553

PR #553, opened 2020-05-29
German "([^ ]+)(strasse|str|straße)" regex parsing

ilissablech:
Add in functionality to apply regex patters to [native/src/text/tokens.rs](https://github.com/mapbox/pt2itp/blob/de_regex/native/src/text/tokens.rs). 

For Germany only, for streets ending in "strasse", "str, or "straße" we will split these endings off into its own word. And then apply the token matching. For example: "banhofstrasse" => "banhof" => "bhf".

In testing a small cluster of addresses this change brought address_orphans down from 1183 to 212 & network_orphans down from 1764 to 801.

### DIFF
--- a/native/src/text/tokens.rs
+++ b/native/src/text/tokens.rs
@@ -1,33 +1,46 @@
-use regex::Regex;
 use super::diacritics;
-use std::collections::HashMap;
 use geocoder_abbreviations::{Token, TokenType};
 use neon::prelude::*;
+use regex::Regex;
+use std::collections::HashMap;
 
 #[derive(Debug, PartialEq, Clone)]
 pub struct Tokens {
-    tokens: HashMap<String, ParsedToken>
+    tokens: HashMap<String, ParsedToken>,
+    regex_tokens: HashMap<String, ParsedToken>,
 }
 
 impl Tokens {
-    pub fn new(tokens: HashMap<String, ParsedToken>) -> Self {
+    pub fn new(tokens: HashMap<String, ParsedToken>, regex_tokens: HashMap<String, ParsedToken>) -> Self {
         Tokens {
-            tokens: tokens
+            tokens: tokens,
+            regex_tokens: regex_tokens,
         }
     }
 
     pub fn generate(languages: Vec<String>) -> Self {
         let import: HashMap<String, Vec<Token>> = geocoder_abbreviations::config(languages).unwrap();
         let mut map: HashMap<String, ParsedToken> = HashMap::new();
+        let mut regex_map: HashMap<String, ParsedToken> = HashMap::new();
 
         for language in import.keys() {
             for group in import.get(language).unwrap() {
-                // if it's a simple, non regex token replacer
                 if !group.regex {
                     for tk in &group.tokens {
                         map.insert(
                             diacritics(&tk.to_lowercase()),
                             ParsedToken::new(diacritics(&group.canonical.to_lowercase()), group.token_type.to_owned())
+                            );
+                        }
+                    }
+                if group.regex {
+                    for tk in &group.tokens {
+                        regex_map.insert(
+                            tk.to_lowercase(),
+                            ParsedToken::new(
+                                group.canonical.to_lowercase(),
+                                group.token_type.to_owned(),
+                            ),
                         );
                     }
                 }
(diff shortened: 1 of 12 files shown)


native/src/text/tokens.rs
```diff
@@ -1,58 +1,104 @@
...
-                // if it's a simple, non regex token replacer
                 if !group.regex {
                     for tk in &group.tokens {
                         map.insert(
                             diacritics(&tk.to_lowercase()),
                             ParsedToken::new(diacritics(&group.canonical.to_lowercase()), group.token_type.to_owned())
+                            );
+                        }
+                    }
+                if group.regex {
```
mattciferri, line 36: [nit] It doesn't seem like the bracket indentation is lining up above this. Also should this be an if/else?

native/src/text/tokens.rs
```diff
@@ -1,58 +1,104 @@
...
+            if country == &String::from("DE") {
+                match self.tokens.get(token) {
+                    None => {
+                        // apply regex before defaulting to tokenized.push(Tokenized::new(token.to_owned(), None))
+                        // loop through regexes to apply any canonical replacements that match
+                        let mut no_token_match = Tokenized::new(token.to_owned(), None);
+                        for (regex_string, v) in self.regex_tokens.iter() {
+                            let re = Regex::new(&format!(r"{}", regex_string));
+                            if re.unwrap().is_match(token) {
+                                let re = Regex::new(&format!(r"{}", regex_string)).unwrap();
```
mattciferri, line 71: Can this be removed? Or the unwrap moved up?

native/src/text/tokens.rs
```diff
@@ -1,58 +1,104 @@
...
+                                let canonical: &str = &*v.canonical; // convert from std::string::String -> &str
+                                let regexed_token = re
+                                    .replace_all(&token, canonical)
+                                    .to_string();
+                                no_token_match =
+                                    Tokenized::new(regexed_token, v.token_type.to_owned());
```
mattciferri, line 77: Is there any risk of us matching multiple regex tokens? Is that a case that we'd want to capture?

native/src/text/tokens.rs
```diff
@@ -1,58 +1,104 @@
...
-                },
-                Some(t) => {
-                    tokenized.push(Tokenized::new(t.canonical.to_owned(), t.token_type.to_owned()));
+            // for right now let's only apply regexes to Germany. English has a negative lookahead query that rust doesn't support.
+            if country == &String::from("DE") {
+                match self.tokens.get(token) {
+                    None => {
+                        // apply regex before defaulting to tokenized.push(Tokenized::new(token.to_owned(), None))
+                        // loop through regexes to apply any canonical replacements that match
+                        let mut no_token_match = Tokenized::new(token.to_owned(), None);
```
mattciferri, line 67: What's the thought behind the naming of `no_token_match`?

native/src/text/tokens.rs
```diff
@@ -65,7 +111,7 @@ impl Tokens {
     /// Remove all diacritics, punctuation non-space whitespace
     /// returning a vector of component tokens
     ///
-    fn tokenize(&self, text: &String) -> Vec<String> {
+    fn tokenize(&self, text: &String, country: &String) -> Vec<String> {
```
mattciferri, line 114: Does this function need the country?

native/src/text/tokens.rs
ilissablech: Line 63 `match self.tokens.get(token)` tries to find a match between our list of tokens to the street name. The `no_token_match` variable is set if the street name has no match. Happy to change the variable name if there is something that would be more clear @mattciferri

native/src/text/tokens.rs
ilissablech: Nope! Removed.

native/src/text/tokens.rs
ilissablech: @mattciferri you're right about the if/else -- fixed. But I don't see the issue with the bracket indentation....

native/src/text/tokens.rs
```diff
@@ -1,58 +1,104 @@
...
                             ParsedToken::new(diacritics(&group.canonical.to_lowercase()), group.token_type.to_owned())
+                            );
+                        }
+                    }
+                if group.regex {
+                    for tk in &group.tokens {
+                        regex_map.insert(
+                            tk.to_lowercase(),
+                            ParsedToken::new(
+                                group.canonical.to_lowercase(),
```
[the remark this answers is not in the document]
ilissablech: @mattciferri that information is not brought into pt2itp. Only `canonical` and `type` are brought in.

native/src/text/tokens.rs
ilissablech: @mattciferri this has only been tested for Germany which has one regex to match on. Should I make a comment in the code to reflect on that as a potential TODO for the future?

mattciferri: @ilissablech Since it would take some testing to get this going for non-German countries, and this would be a case for us to test, maybe just a ticket capturing the current state and next steps?

native/src/text/tokens.rs
mattciferri: What changed here? I still see two instances of `let re = Regex::new(&format!(r"{}", regex_string)).unwrap();`

native/src/text/tokens.rs
mattciferri: Ah I see, I was confused because below we ultimately assign `no_token_match` a matched token. There might be something more clear but I'm good with this :+1:

native/src/text/tokens.rs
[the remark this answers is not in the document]
mattciferri: Right, I'm wondering if it should be. There could be cases similar to this strasse example where we'd want to apply regex, but also strip out diacritics, and that flag would let us know which patht to take. Regardless, it doesn't apply to Germany, so maybe we just make a note of it for the future.

native/src/text/tokens.rs
ilissablech: 🤦 Forgot to push that change. All set.

native/src/text/tokens.rs
ilissablech: Documented in this issue: https://github.com/mapbox/pt2itp/issues/554

native/src/text/tokens.rs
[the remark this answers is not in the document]
ilissablech: Documented in this issue : https://github.com/mapbox/pt2itp/issues/555

native/src/text/tokens.rs
mattciferri: It's all fixed now. The closing bracket above the if was indented too much.

PR #556, opened 2020-08-27
Multi word tokens using spanBoundaries and apostrophe eval

ilissablech:
## Matching on multi-word tokens
We don't currently have the ability to match on multi-word tokens. With this code, we now pull in the "spanBoundaries" property of geocoder abbreviation tokens. The priority for creating the token maps and applying transformations on the names for linking is:
1) regex
2) span boundaries
3) all others

Example: full street name is "gran via de les corts catalanes". The token set for gran via is:
```    
{
    "tokens": [
        "gv",
        "gran via"
    ],
    "full": "gran via",
    "canonical": "gv",
    "spanBoundaries": 1
}
```
"gran via" will be replaced with "gv" and "gv corts catalanes" can be properly matched with "gran via de les corts catalanes".

## Splitting words on apostrophes 
EU  countries have many instances of `l'` and `d'`. The current code removes `'`, making us unable to match the `l'` and `d'` as a type:determiner. 

We now replaces an apostrophe with a space so that  `l'` and `d'` can now match `le` and `de` as a type:determiner.


native/src/text/tokens.rs
```diff
@@ -452,8 +478,7 @@ mod tests {
         assert_eq!(
             tokens.process(&String::from("foo barter"), &String::from("")),
             vec![
-                Tokenized::new(String::from("foo"), None),
-                Tokenized::new(String::from("foo"), None)
```
mattciferri, line 456: How was this yielding two tokens before?

native/src/text/tokens.rs
```diff
@@ -56,59 +79,47 @@ impl Tokens {
...
+                }
+                Some(t) => {
+                    tokenized.push(Tokenized::new(
+                        t.canonical.to_owned(),
+                        t.token_type.to_owned(),
+                    ));
                 }
             }
         }
+        tokenized.dedup(); // dedup
```
mattciferri, line 119: What led to this decision? And can we remove the comment? 😄

native/src/text/tokens.rs
```diff
@@ -26,25 +29,45 @@ impl Tokens {
...
+                        regex_map.insert(
+                            tk.to_lowercase(),
                             ParsedToken::new(
-                                diacritics(&group.canonical.to_lowercase()),
+                                group.canonical.to_lowercase(),
                                 group.token_type.to_owned(),
                             ),
                         );
                     }
+                } else if Some(group.span_boundaries) != Some(None) {
```
mattciferri, line 51: I think there's a better way to do this comparison - `Some(None)` looks a bit funky to me. Does `group.span_boundaries != None` work?

mattciferri (commented):
A few questions but looks good! I'm also curious why there a Cargo.lock and yarn.lock changes but no dependency changes. Did we just run an update?

ilissablech:
@mattciferri thanks for the 👀 . The .lock changes were in there from when I was playing with different commits from geocoder-abbreviations. Removed those. I also removed the dedup call. I has added that in when I was playing with different ways of accomplishing the multi-word replacement goal. Also removed that and fixed the test.

mattciferri (approved):
🌮

PR #557, opened 2020-09-03
Check properties for the orphan flag before selecting them for inclusion in linking

mapbox-danny:
**Context:**
Add a property (interpolate) to addresses in Hecate to flag exclusion from consideration in the interpolation process.
This can be used as a technique to avoid poor conflation outcomes for special case addresses.

**How To Use - interpolate=false:**

To mark an address to be skipped for consideration in interpolation and to be assigned as an address orphan, add the 'interpolate' 'false' tag in JOSM

Screen Shot 2020-09-10 at 2 34 12 PM

### DIFF
--- a/native/src/map/mod.rs
+++ b/native/src/map/mod.rs
@@ -364,7 +364,7 @@ pub fn link_process(conn: &impl postgres::GenericConnection, min: i64, max: i64)
                 match linker::linker(primary, potentials, false) {
                     Some(link_match) => {
                         match trans.execute(&*"
-                            UPDATE address SET netid = $1 WHERE id = $2;
+                            UPDATE address SET netid = $1 WHERE id = $2 AND interpolate = true;
                         ", &[&link_match.id, &id]) {
                             Err(err) => {
                                 println!("Transaction Statement Error: {}", err.to_string());
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Addresses can have any number of name synonyms of equal or differing priority.
 | `street` | `String` or `Array` The name of the street - preferably non-abbreviated. If it's an array, it must contain an object for each street name synonym with the properties `display` for the street name and `priority` for the numeric ranking. |
 | `source` | `String` The source name of the data so a single input file can have a combination of multiple sources |
 | `output` | `Boolean` A boolean allowing pts to be used to calculate the ITP segment but not output in the final cluster |
+| `interpolate` | `Boolean` A boolean, when set to false, keeps the address as an orphan address by skipping its inclusion in the ITP process |
 
 ##### Example
 
--- a/native/src/conflate/mod.rs
+++ b/native/src/conflate/mod.rs
@@ -93,6 +93,7 @@ pub fn conflate(mut cx: FunctionContext) -> JsResult<JsBoolean> {
             number TEXT,
             source TEXT,
             output BOOLEAN,
+            interpolate BOOLEAN,
             props JSONB,
             geom GEOMETRY(POINT, 4326)
         );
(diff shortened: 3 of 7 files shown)


native/src/map/mod.rs
```diff
@@ -400,7 +400,7 @@ pub fn link_process(conn: &impl postgres::GenericConnection, min: i64, max: i64)
                     Some(link_match) => {
                         match trans.execute(
                             &*"
-                            UPDATE address SET netid = $1 WHERE id = $2;
+                            UPDATE address SET netid = $1 WHERE id = $2 AND interpolate = true;
```
bookest, line 403: Would it make sense to move the `interpolate = true` restriction to the initial `SELECT` so that we don't attempt to link any addresses that should be excluded from interpolation? It doesn't look like `linker::linker` has any side effects that we depend on.

mapbox-danny: @bookest, I debated adding the check to the SELECT. Given the likely low number of these flags, my best estimate was that including the check at UPDATE would actually be more performant.

mapbox-danny:
Validation: In staging, I flagged 740 15th St NW, Washington, DC with 'interpolate' as 'false'. I generated an index for DC and validated that in the database the flag correctly came through to the address table. Additionally, a netid was not added to the record (indicating that it had not be matched to the network) and that it was included in the address_orphan_cluster table.

Screen Shot 2020-09-11 at 3 38 10 PM
Screen Shot 2020-09-11 at 3 34 06 PM

I then validated within the index output that the address was returned as an address orphan:

`{"type":"Feature","properties":{"carmen:text":"15th St NW","carmen:addressnumber":[["740"]],"carmen:geocoder_stack":"us","carmen:center":[-77.0339279174805,38.8999252319336],"accuracy":"rooftop","override:postcode":20005},"geometry":{"type":"GeometryCollection","geometries":[{"type":"MultiPoint","coordinates":[[-77.0339279174805,38.8999252319336]]}]},"id":1161831273481940}`

mapbox-danny:
Reviewing the Snapshot: (./mapbox.places-us_dc-address-both.json)
Note similar feature counts based on processing in staging.

STAGING INCLUDING CHANGES:
```    "metrics": {
        "snapshot": "snapshots/mapbox.places-us_dc-address-both-6269b3d51599852332382.geojson",
        "features": {
            "feats": 2759,
            "clusters": 2211,
            "invalid": 0,
            "addresses": 144415,
            "intersections": 19659,
            **"address_orphans": 132**,
            "network_orphans": 416
        }
    },```

STAGING ORIGINAL:
```    "metrics": {
        "snapshot": "snapshots/mapbox.places-us_dc-address-both-3cbda9461599854423344.geojson",
        "features": {
            "feats": 2758,
            "clusters": 2211,
            "invalid": 0,
            "addresses": 144415,
            "intersections": 19659,
            **"address_orphans": 131,**
            "network_orphans": 416
        }
    },```

**Comment by mapbox-danny** [comment]
To validate the Curb fix, I flagged interpolate=false in stage for both 1 and 2 United Nations Plaza.
Below is the address orphan cluster created for those points in the index output: s3://mapbox-places/snapshots/mapbox.places-us_ny-address-both-6269b3d51599854809791.geojson

```
{"type":"Feature","properties":{"carmen:text":"United Nations Plz","carmen:addressnumber":[["1","2"]],"carmen:geocoder_stack":"us","carmen:center":[-73.9697036743164,40.7505760192871],"accuracy":"rooftop","override:postcode":10017},"geometry":{"type":"GeometryCollection","geometries":[{"type":"MultiPoint","coordinates":[[-73.9690780639648,40.7503089904785],[-73.9697036743164,40.7505760192871]]}]},"id":3694255787665102}
```

mattciferri (approved):
🌮

PR #558, opened 2020-09-08
cargo fmt all the things

bookest:
Existing source isn't `cargo fmt` compliant, so we often end up with changes that include unrelated formatting changes due to editors automatically applying formatting fixes. This makes PRs harder to review and makes history less useful as unrelated changes are included in blame output.

This PR does the following:

- runs `cargo fmt` on all existing rust code
- requires all new rust code to have been formatted with `cargo fmt`
- adds `.git-blame-ignore-revs` to allow ignoring these formatting changes in `git blame`

Once this is merged you can configure `git blame` to ignore the formatting changes by running:

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,9 @@ jobs:
                 name: "Cargo Test"
                 command: cd native/ && cargo test --release
             - run:
+                name: "Cargo Format"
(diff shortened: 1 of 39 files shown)


ilissablech (approved):
🌮 yay ty

PR #562, opened 2020-09-17
Subset linking

mapbox-danny:
Add the ability to link based on a subset of tokens
 - all address tokens exist in the network string
or
 - all network tokens exist in the address string

### DIFF
--- a/native/src/types/name.rs
+++ b/native/src/types/name.rs
@@ -296,7 +296,6 @@ impl Name {
         context: &Context,
     ) -> Self {
         let mut display = display.to_string().replace(r#"""#, "");
-        
         // only title case non-generated names
         if source != Some(Source::Generated) {
             display = titlecase(&display, &context);
--- a/native/src/types/name.rs
+++ b/native/src/types/name.rs
@@ -166,8 +166,8 @@ impl Names {
 
     ///
     /// Dedupe a Names struct based on the tokenized version of each name.
-    /// Names with the same priority and tokenized name will preference the dupliacate with the
-    /// longest display name. This tries to prefence non-abbreviated synonyms where they exist,
+    /// Names with the same priority and tokenized name will preference the duplicate with the
+    /// longest display name. This tries to preface non-abbreviated synonyms where they exist,
     /// e.g. 'East Main Street' rather than 'E Main St'
     ///
     pub fn dedupe(&mut self) {
@@ -295,13 +295,12 @@ impl Name {
         source: Option<Source>,
         context: &Context,
     ) -> Self {
-        let mut display = display.to_string().replace(r#"""#, "").replace(r#","#, ""); // commas are not allowed as they are used to delimit synonyms on output
+        let mut display = display.to_string().replace(r#"""#, ""); // commas are not allowed as they are used to delimit synonyms on output
 
         // only title case non-generated names
         if source != Some(Source::Generated) {
             display = titlecase(&display, &context);
         }
-
         let tokenized = context.tokens.process(&display, &context.country);
 
         if context.country == String::from("US") || context.country == String::from("CA") {
@@ -488,7 +487,7 @@ mod tests {
         assert_eq!(
             Name::new(String::from(","), 0, None, &context),
             Name {
-                display: String::from(""),
+                display: String::from(","),
                 priority: 0,
                 source: None,
                 tokenized: vec![],
--- a/native/src/types/name.rs
+++ b/native/src/types/name.rs
@@ -295,8 +295,8 @@ impl Name {
         source: Option<Source>,
         context: &Context,
     ) -> Self {
-        let mut display = display.to_string().replace(r#"""#, ""); // commas are not allowed as they are used to delimit synonyms on output
-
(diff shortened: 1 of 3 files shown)


native/src/util/linker.rs
```diff
@@ -176,14 +171,77 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
+                // Don't bother considering if the tokenless forms don't share a starting letter
+                // this might require adjustment for countries with addresses that have leading tokens
+                // which aren't properly stripped from the token list
+                if subset_match
+                    || (potential_tokenless.len() > 0
+                        && tokenless.len() > 0
+                        && potential_tokenless.get(0..1) != tokenless.get(0..1))
+                {
+                    continue;
+                }
```
bookest, line 244: Is this short circuit still necessary? It looks like it exists to short circuit to the next potential match, but it's been moved to the end of the loop, so the `continue` doesn't do anything here.

native/src/util/linker.rs
```diff
@@ -176,14 +171,77 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
+                    for ntok in &ntoks {
+                        // Check if all tokens in the network are preset within the address
+                        let atok_index = &atoks.iter().position(|r| r == ntok);
+                        if (atok_index.is_none()) {
+                            network_subset_match = false;
+                        } else {
+                            atoks.remove(atok_index.unwrap());
+                        }
+                    }
+
```
bookest, line 223: It looks like the implementation doesn't care about order, in which case this check can be performed more simply using `std::collections::HashSet`s and `is_subset`. Though I'm unsure if the order independent comparison is the intent or if we mean to be testing that the address tokens are a subsequence of the network tokens (and vice versa).

native/src/util/linker.rs
```diff
@@ -176,14 +171,77 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
+                        .map(|x| x.token.to_owned())
+                        .collect();
+
+                    let mut address_subset_match = true;
+                    let mut network_subset_match = true;
+
+                    for atok in &atoks {
+                        // Check if all tokens in the address are present within the network
+                        let ntok_index = &ntoks.iter().position(|r| r == atok);
+                        if (ntok_index.is_none()) {
```
mattciferri, line 201: Can this be made a match statement to avoid the unwrap?

native/src/util/linker.rs
```diff
@@ -131,6 +123,9 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
                 // Use a weighted average w/ the tokenless dist score if possible
                 let mut lev_score: Option<f64> = None;
 
+                let tokenlev = distance(&tokenized, &potential_tokenized) as f64;
+                let tokenlesslev = distance(&tokenless, &potential_tokenless) as f64;
```
mattciferri, line 127: Are these used?

native/src/util/linker.rs
```diff
@@ -176,14 +171,77 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
                     - (((2.0 * lev_score.unwrap())
                         / (potential_tokenized.len() as f64 + tokenized.len() as f64))
                         * 100.0);
 
+                // check for subset matches, overriding scores below the matching criteria
+                if (score <= 70.0
+                    && tokenized.len() >= 2
+                    && potential_tokenized.len() >= 2
+                    && tokenless.len() >= 1
+                    && potential_tokenless.len() >= 1)
```
mattciferri, line 184: Is there a reason tokenless >=1 is required?

mattciferri (changes requested):

native/src/types/name.rs
```diff
@@ -295,13 +295,12 @@ impl Name {
         source: Option<Source>,
         context: &Context,
     ) -> Self {
-        let mut display = display.to_string().replace(r#"""#, "").replace(r#","#, ""); // commas are not allowed as they are used to delimit synonyms on output
+        let mut display = display.to_string().replace(r#"""#, ""); // commas are not allowed as they are used to delimit synonyms on output
```
bookest, line 298: Should this comment be removed since we're no longer stripping out commas here?

bookest (approved):
A couple of places where we can short circuit to avoid doing some work once we know we either are (or are not) a submatch, but otherwise LGTM.

native/src/types/name.rs
mattciferri: @ilissablech Why are we no longer removing commas?

ilissablech: @mattciferri items like `Rue Du Dix Neuf Mars 1962,D 149.10` were becoming `Rue Du Dix Neuf Mars 1962D 149.10` & we have a spot to remove commas later in the code that does not interfere with this type of situation https://github.com/mapbox/pt2itp/blob/master/native/src/text/tokens.rs#L130

PR #565, opened 2020-10-09
Substring linking abbreviations

mapbox-danny:
Check for abbreviations within substring matching

An abbreviation is defined as:
- begins with the same character as the full word
- all characters are found within the full word, in order


ilissablech:
Closing PR - resolved with https://github.com/mapbox/pt2itp/pull/571

PR #566, opened 2020-11-10
Allow house number slashes for Czech Republic and Poland

ilissablech:
Czech Republic and Poland both have house numbers formatted as `123/45`. We want to allow that format through for only Czech Republic and Poland. Still, the format of `123 1/2` will not be acceptable.

### DIFF
--- a/native/src/types/address.rs
+++ b/native/src/types/address.rs
@@ -117,7 +117,8 @@ impl Address {
             geom: geom,
         };
 
-        addr.std()?;
+        let country = &context.country.to_lowercase();
+        addr.std(country)?;
 
         Ok(addr)
     }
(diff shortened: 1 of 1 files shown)


native/src/types/address.rs
```diff
@@ -229,15 +229,35 @@ impl Address {
...
-        if !SUPPORTED.is_match(self.number.as_str()) {
-            return Err(String::from("Number is not a supported address/unit type"));
+        // Czech Republic and Poland have addresses in the format of "123/89"
+        // Let's allow those through, but still not 123 1/2, regardless of country
+        if SLASH_EXCLUDED_COUNTRIES.contains(&country) {
+            if !SLASH_SUPPORTED.is_match(self.number.as_str()) {
+                return Err(String::from("Number is not a supported address/unit type"));
+            }
+        } else {
+            if !SUPPORTED.is_match(self.number.as_str()) {
```
bookest, line 258: It might make sense to rename `SUPPORTED` to something like `DEFAULT_SUPPORTED`.

native/src/types/address.rs
```diff
@@ -229,15 +229,35 @@ impl Address {
...
+                return Err(String::from("Number is not a supported address/unit type"));
+            }
```
bookest, line 260: I think you can remove the duplicated `if is_match` using something like:

```rust
let supported = if SLASH_EXCLUDED_COUNTRIES.contains(&country) { 
    &SLASH_SUPPORTED 
} else { 
    &SUPPORTED 
};
if !supported.is_match(self.number.as_str()) {
    return Err(String::from("Number is not a supported address/unit type"));
}
```

native/src/types/address.rs
```diff
@@ -535,5 +555,37 @@ mod tests {
...
+            );
+
+            let addr = Address::new(feat, &context).unwrap();
+
+            assert_eq!(addr.to_tsv(), "\t0\t[{\"display\":\"Hickory Hills Dr\",\"priority\":-1,\"source\":\"Address\",\"tokenized\":[{\"token\":\"hickory\",\"token_type\":null},{\"token\":\"hills\",\"token_type\":null},{\"token\":\"dr\",\"token_type\":null}],\"freq\":1}]\t123/89\tTIGER-2016\tfalse\ttrue\t{\"number\":\"123/89\",\"source\":\"TIGER-2016\",\"street\":\"Hickory Hills Dr\"}\t0101000020E6100000096C0B88B40D55C00BF02796EB9B4340\n");
+        }
+    }
+
+    #[test]
+    #[should_panic(expected = "Number is not a supported address/unit type")]
```
bookest, line 576: Does this need to panic? Could this check be implemented in the test itself by checking the `Error` value directly instead of calling `unwrap()`?

bookest (approved):
A couple of suggestions, but LGTM.

native/src/types/address.rs
ilissablech: @bookest that format returns an incompatible type error:
 ```
`if` and `else` have incompatible types

expected struct `types::address::Address::std::SLASH_SUPPORTED`, found struct `types::address::Address::std::DEFAULT_SUPPORTED`

note:   expected type `&types::address::Address::std::SLASH_SUPPORTED`
```

PR #567, opened 2020-11-12
Skip lookahead token replacement for the UK

aarthykc:
## Context

The process to load UK addresses crashed while attempting run token replacement. This is because we're using an english token list for token replacement, which contains complex token replacements. The loading processed crashed with the following error: 
```
---- types::name::tests::test_tokenized_string stdout ----
thread 'types::name::tests::test_tokenized_string' panicked at 'called `Result::unwrap()` on an `Err` value: Syntax(
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    ([redacted])
     ^^^
error: look-around, including look-ahead and look-behind, is not supported
```
We have some logic in place right now that gets around the error, and this PR extends that logic to the UK. There's probably a better approach that doesn't require the developer to do anything additional when including english tokens - but given the rushed timelines probably fine to add it to a list.

cc @mapbox/addresses


ilissablech (approved):
LGTM

PR #568, opened 2020-11-16
Update to geocoder-abbreviations v4.5.0

bookest:
This bumps geocoder-abbreviations on both the Rust and JavaScript sides. The JavaScript side hasn't been updated in quite a while and requires updating a test to account for changes in the global tokens provided by geocoder-abbreviations.


ilissablech (approved):
LGTM

PR #571, opened 2020-12-10
Abbreviation matching

ilissablech:
In a [previous PR](https://github.com/mapbox/pt2itp/pull/562) we added logic to add the ability to link based on a subset of tokens. This PR extends that, adding the ability to match on abbreviations.

Examples have been added in the test suite:
- M. Pišúta will match with Milana Pišúta
- A. Kostolného will match with Andreja Kostolného Kostolného
- Bo Ntra will match with Barrio Nuestra

Also added a bunch of tests and a note in the README about how to run Cargo tests.

Notes on address orphan differences for various EU countries: https://github.com/mapbox/mapbox-places-address/issues/1196

cc: @mapbox/search-addresses @flatwhite0928

### DIFF
--- a/native/src/util/linker.rs
+++ b/native/src/util/linker.rs
@@ -1419,4 +1419,54 @@ mod tests {
             assert_eq!(linker(a, b, false), Some(LinkResult::new(2, 70.01)));
         }
     }
+    #[test]
+    fn test_it_linker() {
+        let mut tokens: HashMap<String, ParsedToken> = HashMap::new();
+        let regex_tokens: HashMap<String, ParsedToken> = HashMap::new();
+        let context = Context::new(
+            String::from("it"),
+            None,
+            Tokens::generate(vec![String::from("it")]),
+        );
+        {
+            let a_name = Names::new(vec![Name::new("Via Angelo Silvio Novaro", 0, None, &context)], &context);
+            let b_name = Names::new(
+                vec![Name::new("Via A. S. Novaro", 0, None, &context)],
+                &context,
+            );
+            let a = Link::new(1, &a_name);
+            let b = vec![Link::new(2, &b_name)];
+            assert_eq!(linker(a, b, false), Some(LinkResult::new(2, 70.01)));
+        }
+    }
+    #[test]
+    fn test_be_linker() {
+        let mut tokens: HashMap<String, ParsedToken> = HashMap::new();
+        let regex_tokens: HashMap<String, ParsedToken> = HashMap::new();
+        let context = Context::new(
+            String::from("fr"),
+            None,
+            Tokens::generate(vec![String::from("fr")]),
+        );
+        {
+            let a_name = Names::new(vec![Name::new("rue de la reine astrid", 0, None, &context)], &context);
+            let b_name = Names::new(
+                vec![Name::new("rue reine astrid", 0, None, &context)],
+                &context,
+            );
+            let a = Link::new(1, &a_name);
+            let b = vec![Link::new(2, &b_name)];
+            assert_eq!(linker(a, b, false), Some(LinkResult::new(2, 91.18)));
+        }
+        {
+            let a_name = Names::new(vec![Name::new("grand'place", 0, None, &context)], &context);
+            let b_name = Names::new(
+                vec![Name::new("grand place", 0, None, &context)],
+                &context,
+            );
+            let a = Link::new(1, &a_name);
+            let b = vec![Link::new(2, &b_name)];
+            assert_eq!(linker(a, b, false), Some(LinkResult::new(2, 70.01)));
+        }
+    }
 }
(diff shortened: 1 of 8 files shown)


native/src/util/linker.rs
```diff
@@ -35,6 +35,67 @@ impl LinkResult {
...
+            None => {
+                if token_list_two.len() > 0 {
+                    // if abbreviation is found return stripped through abbreviated index value
+                    let check_abbreviation: (bool, Vec<String>) = is_abbrev(word, &token_list_two);
+                    abbrev_match = check_abbreviation.0;
+                    token_list_two = check_abbreviation.1;
+                }
+                if !abbrev_match {
+                    return false;
+                }
```
bookest, line 92: The intent behind `abbrev_match` is a little hard to follow here. It looks like it can be moved inside the `if token_list_two.len() > 0` block without affecting the flow here. Is that correct?

```rust
None => {
    if !token_list_two.is_empty() {
        let (abbrev_match, remaining_tokens) = is_abbrev(word, &token_list_two);
        // do we rely on token_list_two being updated if we're returning false?
        token_list_two = remaining_tokens;
        if !abbrev_match {
            return false;
        }
    }
}
```

package.json
```diff
@@ -16,6 +16,8 @@
     "coverage-upload": "coveralls < ./coverage.lcov",
     "pretest": "test/pretest.js",
     "test": "tape test/*.test.js && cd native && cargo test --release",
+    "cargo_all": "cd native && cargo test --release",
+    "cargo_one": "cd native && cargo test --release -- --exact util::linker::tests::test_fr_linker -- --nocapture",
```
bookest, line 20: Is there something about this specific test that it shoudl have a hardcoded shortcut? I think it would be more useful if this script took the test to run as an argument.

native/src/util/linker.rs
```diff
@@ -1228,5 +1301,172 @@ mod tests {
...
+                &context,
+            );
+            let b_name = Names::new(
+                vec![Name::new("cl federico garcia lorca", 0, None, &context)],
+                &context,
+            );
+            let a = Link::new(1, &a_name);
+            let b = vec![Link::new(2, &b_name)];
+            assert_eq!(linker(a, b, false), Some(LinkResult::new(2, 70.01)));
+        }
```
bookest, line 1316: [nit] This pattern is repeated several times throughout the tests. Refactoring it out to a function (or a custom assert macro) would help cut down on test boilerplate.

bookest (approved):
I left a couple of inline comments, but overall LGTM.

bookest:
Looks like the build is failing on the `cargo fmt` check.

native/src/util/linker.rs
```diff
@@ -267,6 +297,44 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
+            String::from($language),
+            None,
+            Tokens::generate(vec![String::from($language)]),
+        );
+    };
+}
+
+#[macro_export]
+macro_rules! assert_linker_eq {
+    ($language:expr; $name_a:expr; $name_b:expr; $strict_mode:expr; $expected_return:expr) => {
```
bookest, line 313: Does this need to use `;` as the separator here? I think it can use `,` to read more like a function and match other assert macros.

native/src/util/linker.rs
```diff
@@ -267,6 +297,44 @@ pub fn linker(primary: Link, mut potentials: Vec<Link>, strict: bool) -> Option<
...
+
+#[macro_export]
+macro_rules! assert_linker_no_match {
+    ($language:expr; $name_a:expr; $name_b:expr; $strict_mode:expr) => {
+        let context = build_lang_context!($language);
+        let a_name = Names::new(vec![Name::new($name_a, 0, None, &context)], &context);
+        let b_name = Names::new(vec![Name::new($name_b, 0, None, &context)], &context);
+        let a = Link::new(1, &a_name);
+        let b = vec![Link::new(2, &b_name)];
+        assert_eq!(linker(a, b, false), None)
```
bookest, line 334: This should use `$strict_mode` instead of hard coding `false`, right?

native/src/util/linker.rs
```diff
@@ -1058,175 +1087,106 @@ mod tests {
...
-                &context,
-            );
-            let a = Link::new(1, &a_name);
-            let b = vec![Link::new(2, &b_name)];
-            assert_eq!(linker(a, b, false), None);
+            assert_linker_eq!("fr"; "rue de l'eglise"; "saint martin ruet de l'eglise encore"; false; 70.01);
+        }
+        {
+            assert_linker_eq!("fr"; "rue de l'eglise saint martin"; "rue de l'eglise"; false; 70.01);
+        }
```
bookest, line 1115: I think the need for the new scope for each `assert_linker_eq!` call is a signal that the macro should be creating the scope. Otherwise it's easy for the user to forget and potentially conflict with existing names.

bookest (approved):

bookest:
Nice job on the custom assertions. They really clean up the linker tests quite a bit.

PR #573, opened 2021-01-25
Hyphen match

ilissablech:
New York and Hawaii have addresses with hyphens (ie. `43-27 247th st`). Some address providers include the hyphen while others do not (ie. `4327 247 st`). We need to be able to find an address match if a) the address is an exact match OR b) the address with hyphens removed matches.

cc: @mapbox/search-addresses


bookest (approved):
LGTM

ilissablech:
Closing as now handled in the geodata pipelines.

PR #574, opened 2021-02-03
Add Canada to skip_regex_list

ilissablech:
Add Canada to the list of countries that are using English tokens and regex should not be applied.


bookest:
What's the impact on Canadian addresses in French?

bookest (approved):
LGTM

PR #575, opened 2021-04-30
Ignore lookahead token replacement errors Ireland

ilissablech:
Skip look ahead regex for Ireland.

Upgrade the version of geocoder-abbreviations to pull in tokens for Latvia and Bulgaria.


mapbox-danny (approved):
Looks good

PR #582, opened 2021-09-27
Skip negative lookahead for AU, NZ and bump version

flatwhite0928:
### Background
https://github.com/mapbox/mapbox-places-address/pull/1312
This PR is going to skip negative lookahead for AU, NZ, because they use EN tokens, and negative lookahead is just for US. Bump the version of pt2itp.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ## Version History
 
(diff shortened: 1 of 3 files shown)


ilissablech (approved):
LGTM

PR #587, opened 2022-04-26
Upgrade to Node 14

ianshward:
This PR upgrades pt2itp to Node 14, including several dependencies that were required to be upgraded in order to upgrade to Node 14.  A couple of call-outs:

-  In order for tests to run both locally and on CircleCI, I had to reduce the precision of some of the lat/lon coordinates in fixtures. Values in test "actual" would get truncated for me locally, but I did not get to the bottom of what exactly about my local setup would cause this. However, the reduced precision is still highly precise.
- This PR also upgrades Carmen, and something appears to have changed in the Carmen debug server output (but seems undocumented in the Carmen changelog), but it was easy enough to adapt the tests.

Todo:

- [x] Update error messages in Rust bindings
- [x] Update CHANGELOG
- [x] Before merging, update version in package.json
- [ ] Tag & push new tag


native/Cargo.toml
```diff
@@ -20,8 +20,8 @@ regex = "1.1"
 rstar = "0.4"
 postgis = "0.6.0"
 geo = "0.12.2"
-neon = "0.2"
-neon-serde = "0.1.1"
+neon = "0.7.1"
+neon-serde = { git = "https://github.com/mattciferri/neon-serde.git", rev="a9d9164f45ff65f6f64037fe8cb3490bed357580" }
```
ilissablech, line 24: Should we keep it like this? Is there any worry that Matt might remove this?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/pt2itp",
-  "version": "28.8.3",
+  "version": "29.0.0-dev3",
```
ilissablech, line 3: Are we going to bump to a non-dev version number?

ilissablech (approved):
LGTM. Only two questions on versions.

PR #589, opened 2022-05-03
Run cargofmt on master

ianshward:
For some reason this step passed in an earlier build, so I'm not sure why it didn't fail until after merging.  In any case, this fixes format/style issues in the rust changes on the Node 14 branch that was just merged to master.


ilissablech (approved):
LGTM

PR #591, opened 2022-05-05
Fix name of command in readme (cargo_individual not cargo_one)

ianshward:
This fixes a typo/misnaming in the example of how to run a single test in cargo using cargo_individual


ianshward:
It also looks like `--my_test` does not work, and it should just be `npm run cargo_individual types::name::tests::test_names` for example

PR #593, opened 2022-05-11
Remove low frequency / outlier names coming from address point source

ianshward:
This PR adds a filter_outliers function which removes names from address clusters when those names are of low frequency and from address point source/data

## Next steps

- [x] Update changelog w/ new tag name
- [x] Update version in package.json
- [ ] Review + merge
- [ ] Git tag 29.1.0
- [ ] NPM publish

### DIFF
--- a/native/src/types/name.rs
+++ b/native/src/types/name.rs
@@ -197,10 +197,13 @@ impl Names {
                         || name.display.len() > d.name.display.len()
                     {
                         let priority = d.name.priority;
-                        let freq = d.name.freq;
+                        let freq = d.name.freq + name.freq;
                         d.name = name;
                         d.name.priority = priority;
                         d.name.freq = freq;
+                    } else {
+                        // Keep track of frequency
+                        d.name.freq = d.name.freq + name.freq;
                     }
                 }
                 // if it doesn't yet exist, add it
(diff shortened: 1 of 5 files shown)


flatwhite0928 (approved):
LGTM

native/src/types/name.rs
```diff
@@ -235,6 +238,27 @@ impl Names {
...
+    pub fn filter_outliers(&mut self) {
+        let total_freq: i64 = self.names.iter().map(|name| name.freq).sum();
+        // Only filter when there are more than 10 addresses in the cluster
+        if total_freq > 10 {
+            let mut temp_names: Vec<Name> = self.names.clone();
+            // Remove names that represent 8% or less of names in the cluster
+            // and only if their source is from an address point
+            temp_names.retain(|name| {
+                name.source != Some(Source::Address)
+                    || (name.freq as f32 / total_freq as f32) > 0.08
```
mapbox-danny, line 253: What do you think about a hard limit here as well? 
Say for example that you must have at least 2-3 points matching a name for it to be included (when another option is available)?
Conceptually we might run into an issue in an area with low coverage or a short road segment with 1 of 12 or less bad points (0.0833).

mapbox-danny (approved):
Approved with a suggestion for a potential future addition

